Issue #2, opened 2020-07-21
NOOP untagged response sent by Zimbra


c-leroy:
Hello,

I quite often encounter a response parsing error while issuing long running requests to our Zimbra server :
```
Traceback (most recent call last):
	8: from .rvm/rubies/ruby-2.6.5/lib/ruby/2.6.0/net/imap.rb:1120:in `block in initialize'
	7: from .rvm/rubies/ruby-2.6.5/lib/ruby/2.6.0/net/imap.rb:1147:in `receive_responses'
	6: from .rvm/rubies/ruby-2.6.5/lib/ruby/2.6.0/net/imap.rb:1244:in `get_response'
	5: from .rvm/rubies/ruby-2.6.5/lib/ruby/2.6.0/net/imap.rb:2181:in `parse'
	4: from .rvm/rubies/ruby-2.6.5/lib/ruby/2.6.0/net/imap.rb:2255:in `response'
	3: from .rvm/rubies/ruby-2.6.5/lib/ruby/2.6.0/net/imap.rb:2308:in `response_untagged'
	2: from .rvm/rubies/ruby-2.6.5/lib/ruby/2.6.0/net/imap.rb:2879:in `text_response'
	1: from .rvm/rubies/ruby-2.6.5/lib/ruby/2.6.0/net/imap.rb:3347:in `match'
.rvm/rubies/ruby-2.6.5/lib/ruby/2.6.0/net/imap.rb:3495:in `parse_error': unexpected token CRLF (expected SPACE) (Net::IMAP::ResponseParseError)
```
The last time, the server sent a untagged response "* NOOP" while a `close` request was in progress, after I had flagged lots of mails as Deleted :
```
[...]
C: RUBY0037 UID STORE 3668787,3668788,3668790,3668791 +FLAGS (\Deleted)
S: * 3480 FETCH (FLAGS (\Deleted) UID 3668787)
S: * 3481 FETCH (FLAGS (\Deleted) UID 3668788)
S: * 3482 FETCH (FLAGS (\Deleted) UID 3668790)
S: * 3483 FETCH (FLAGS (\Deleted) UID 3668791)
S: RUBY0037 OK UID STORE completed
C: RUBY0038 CLOSE
S: * NOOP
@str: "* NOOP\r\n"
@pos: 8
@lex_state: EXPR_BEG
@token.symbol: CRLF
@token.value: "\r\n"
[backtrace]
```
[...]

PR #3, opened 2020-07-24
Add Net::IMAP::IgnoredResponse

c-leroy:
Zimbra server can send an untagged response "* NOOP\r\n" in order to
keep alive connection during long-running requests. These responses are
not syntactically correct, but can be safely ignored.

Add Net::IMAP::IgnoredResponse, and use it for this case.

Fixes ruby/net-imap#2


c-leroy:
I have opened a ticket on Ruby redmine : https://bugs.ruby-lang.org/issues/17049

shugo:
@c-leroy Thank you.  It looks fine and I've merged it.

PR #4, opened 2020-11-06
net/imap: support IMAP4 ID extension (RFC2971)

nevans:
IMAP servers often strongly recommend clients to report their ID, for
usage statistics and bug reports. Clients may also find server IDs
useful for the same reasons.

Although the RFC requires otherwise, some servers do make it a soft
requirement that the ID is reported (and will sometimes even disconnect
clients they do not recognize).

See https://tools.ietf.org/html/rfc2971 for more details.

n.b. I submitted this PR previously, before net-imap was gemified:
 * https://bugs.ruby-lang.org/issues/16610
 * https://github.com/ruby/ruby/pull/2882

### DIFF
--- a/lib/net/imap.rb
+++ b/lib/net/imap.rb
@@ -385,7 +385,7 @@ module Net
     def id(client_id=nil)
       synchronize do
         send_command("ID", ClientID.new(client_id))
-        return @responses.delete("ID")[-1]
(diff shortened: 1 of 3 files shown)


shugo:
@nevans Thank you.  I've merged it manually.

PR #7, opened 2020-11-06
Extract public Net::IMAP.authenticator

nevans:
Other IMAP libraries (e.g. using EventMachine or async-io) may want to
reuse some of Net::IMAP's code. This allows libraries to access and share
Net::IMAP's authenticator registry.


shugo:
@nevans I've merged it.  Thank you!

PR #8, opened 2020-11-07
Fix text_response and resp-text

nevans:
This fixes e.g.:
```
RUBY0004 OK [READ-WRITE] [Gmail]/Sent Mail selected. (Success)
```

Specifically, "text" is allowed to begin with "[" or "=". Excluding those was a an older RFC 2060 restriction.

(RFC 2060 was written in 1996. RFC 3501 in 2003)  :slightly_smiling_face:

PR #9, opened 2020-11-09
add BADCHARSET support

nevans:
I was having a hard time extracting this one from my codebase without creating lots of merge conflicts. So instead of making it narrowly self-contained, I built it on top of #6 and #8.  The entire unique PR is contained in 0925990, and everything before that commit belongs to those other two PRs.


shugo:
@nevans I've merged it.  Thank you!

PR #21, opened 2021-04-26
Update Gemfile.lock with newer versions

nevans:
`rake` could already run without bundler, but this allows `bundle exec rake` to work with ruby 2.6, 2.7, and 3.0.
Otherwise, those versions all fail with something like the following error:

    You have already activated power_assert 1.1.7, but your Gemfile
    requires power_assert 1.1.5. Prepending `bundle exec` to your
    command may solve this. (Gem::LoadError)

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -1,14 +1,20 @@
 PATH
   remote: .
   specs:
-    net-imap (0.1.0)
+    net-imap (0.2.1)
+      digest
+      net-protocol
+      strscan
 
 GEM
   remote: https://rubygems.org/
   specs:
-    power_assert (1.1.5)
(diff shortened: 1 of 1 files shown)


shugo:
I'd like to remove Gemfile.lock instead.
@hsbt Do you have any considerations?

hsbt:
@shugo agreed. The existence of `Gemfile.lock` is my laziness. I removed it at https://github.com/ruby/net-imap/commit/84e833b7640de0c9081ae7d4043170c11a4f94b3

shugo:
@hsbt Thank you!

nevans:
@hsbt great! thanks. :)

Issue #10, opened 2020-11-14
Split out response parser (etc) into multiple files


nevans:
I would find it much easier to work with if net/imap.rb were split into several files. Would a patch like that be welcome and accepted? Just let me know, and I'll submit a PR right away (and if it's accepted, I'll rebase all existing PRs on it).

e.g:
- `net/imap/response_parser.rb` for `Net::IMAP::ResponseParser`
- `net/imap/response_data.rb` for the response data Structs (e.g. `TaggedResponse`, `FetchData`, `MailboxList`, etc)
- `net/imap/command_data.rb` for all of the `#send_data`/`#validate classes` (like `Atom`, `Literal`, `QuotedString`, etc)
- `net/imap/authenticators.rb`
  - `net/imap/authenticators/{cram_md5,login,plain}` etc

nevans:
@shugo Other than the minor changes mentioned in #22 and a bunch of documentation updates (see 9cd562a), the code has simply been moved to different files without any changes. It matched my proposal here so I pushed them directly. Let me know if you think we ought to organize anything differently. 🙂

PR #22, opened 2021-04-28
Extract authenticators to their own files

nevans:
I've started down the path of implementing #10. This PR simply copies authenticators with some minor changes, but this raises a big issue I've been meaning to ask about: SASL.  I'll open another issue to discuss that: #23.  I'm marking this as a draft, pending that discussion.

In addition to moving the authenticators to new files, I've updated the rdoc and added `authnid` support for `PLAIN`.

### DIFF
--- a/lib/net/imap.rb
+++ b/lib/net/imap.rb
@@ -16,8 +16,6 @@
 
 require "socket"
 require "monitor"
-require "digest/md5"
-require "strscan"
 require 'net/protocol'
 begin
   require "openssl"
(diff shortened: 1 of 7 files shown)


shugo (approved):
It looks good to me.
@hsbt Do you have any considerations in terms of Ruby releases?

hsbt:
👍  It seems good practice for organizing code.

@nevans Can you also move `net-imap.gemspec` to under the `net/imap` dir?

nevans:
> @nevans Can you also move `net-imap.gemspec` to under the `net/imap` dir?

If we move the gemspec, won't that break existing build scripts?  In particular, it means that people won't be able to use the common pattern: `gem 'net-imap', github: "ruby/net-imap", ref: ...`. They could add `glob: 'lib/net/imap/*.gemspec'` but that's a fairly uncommon option, especially when there's only one gem in the repo.

nevans:
And... I only just discovered that [bundler's default glob](https://github.com/rubygems/rubygems/blob/4f596163feb6fe1c7469e18d64c9b56bd71e032e/bundler/lib/bundler/source/path.rb#L14) is `"{,*,*/*}.gemspec"`.  I'd assumed it was different, matching less.  Still, that means the gemfile would need to be in the repo root or one dir down. Neither `lib/net/imap/net-imap.gemspec`  nor `net/imap/net-imap.gemspec` would work.  Am I misunderstanding something? 🙂

hsbt:
> f we move the gemspec, won't that break existing build scripts?

Ah, Sorry. I meant it in `ruby/ruby` repo. There is no action with this repo.

hsbt:
I tweaked sync script at https://github.com/ruby/ruby/commit/5de6f1ab4751f02670239ebe537e599c477dc4bf

PR #26, opened 2021-11-18
Remove max_flag_count. Ruby 2.2+ can GC symbols.

nevans:
Removing this isn't a high priority, but...

I believe the original reason for `max_flag_count` was because of GC concerns with uncollected symbols.  However, ruby can GC symbols now.  Any unbounded memory consumption or memory leaks most likely comes from other culprits (e.g. accumulating unhandled responses).

The `max_flag_count` was added in 5049da5, 2009-11-19, during the ruby 1.9.1 series.

@shugo  What do you think?


shugo (approved):
It looks fine.  Thank you!

PR #27, opened 2021-11-18
Move NumValidator and Errors to other files

nevans:
Just a few more moves of various classes, related to #10.  Similar to c2408aa, 746757b, 24e929f, 0d43c5e, 2a9afa8, 64d1080.

All of the exception classes were moved into `errors.rb`.  And I moved `NumValidator` into `data_encoding.rb`.  `NumValidator` is currently only used internally when constructing commands, but it can also be used to validate responses (e.g. I've seen a couple of servers that attempt to send 64bit `UIDVALIDITY`! 😖)

### DIFF
--- a/lib/net/imap.rb
+++ b/lib/net/imap.rb
@@ -13,7 +13,6 @@
 # See Net::IMAP for documentation.
 #
 
-
 require "socket"
 require "monitor"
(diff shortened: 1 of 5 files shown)


shugo (approved):
It looks fine.  Thank you!

PR #28, opened 2021-11-19
Add and document flags from RFC9051

nevans:
n.b. https://github.com/ruby/rdoc/issues/345 means that the `:section:` rdoc comments won't show up in the HTML rdoc output. 🙁


shugo (approved):
It looks fine.  Thank you!

Issue #55, opened 2022-01-09
RFC6331: Moving DIGEST-MD5 to Historic


Neustradamus:
Dear @ruby team,

In first, I wish you a Happy New Year!

20 November 2008: CRAM-MD5 to Historic:
- https://tools.ietf.org/html/draft-ietf-sasl-crammd5-to-historic-00

29 June 2017: CRAM-MD5 to Historic:
- https://tools.ietf.org/html/draft-zeilenga-luis140219-crammd5-to-historic-00

July 2011: RFC6331: Moving DIGEST-MD5 to Historic:
- https://tools.ietf.org/html/rfc6331

August 2021: RFC9051: Internet Message Access Protocol (IMAP) - Version 4rev2:
"Replaced DIGEST-MD5 SASL mechanism with SCRAM-SHA-256. DIGEST-MD5 was deprecated."
- https://tools.ietf.org/html/rfc9051

I add same about SCRAM-MD5.

There are now:
- July 2010: RFC5802: Salted Challenge Response Authentication Mechanism (SCRAM): SASL and GSS-API Mechanisms: https://tools.ietf.org/html/rfc5802 (SCRAM-SHA-1 and SCRAM-SHA-1-PLUS)
- July 2010: RFC5803: Lightweight Directory Access Protocol (LDAP) Schema for Storing Salted: Challenge Response Authentication Mechanism (SCRAM) Secrets: https://tools.ietf.org/html/rfc5803
- November 2015: RFC7677: SCRAM-SHA-256 and SCRAM-SHA-256-PLUS: Simple Authentication and Security Layer (SASL) Mechanisms: https://tools.ietf.org/html/rfc7677

Soon:
- SCRAM-SHA-512(-PLUS): https://tools.ietf.org/html/draft-melnikov-scram-sha-512
- SCRAM-SHA3-512(-PLUS): https://tools.ietf.org/html/draft-melnikov-scram-sha3-512

nevans:
Thank you so much for simply documenting this so thoroughly and cleanly, @Neustradamus.

100% Agreed. I believe we these mechanisms should be available but not loaded by default. In order to use them, users would need to do the following:

```ruby
# if SASL remains in this repository as-is:
require "net/imap/authenticators/deprecated/cram-md5_authenticator"
require "net/imap/authenticators/deprecated/digest-md5_authenticator"
require "net/imap/authenticators/deprecated/login_authenticator"
```

[...]

PR #56, opened 2022-02-02
Remove needless dependencies

eregon:
These are default gems, so there is no need to explicitly depend on them,
and depending on them is actually harmful: https://bugs.ruby-lang.org/issues/18567.

### DIFF
--- a/net-imap.gemspec
+++ b/net-imap.gemspec
@@ -32,6 +32,4 @@ Gem::Specification.new do |spec|
   spec.require_paths = ["lib"]
 
   spec.add_dependency "net-protocol"
-  spec.add_dependency "digest"
-  spec.add_dependency "strscan"
 end


hsbt:
I think we need to remove only `net-protocol` because digest and strscan may remove completely from stdlib in the future. But `net-protocol` is not.

eregon:
Ah, I thought `net-protocol` is an external non-stdlib gem too, but indeed it's a default gem.
`net-protocol` is no problem anyway.

I don't see `strscan` or `digest` being removed from stdlib anytime soon/ever (so many programs depend on them, and adding the explicit dependency is annoying, especially for small scripts without Bundler). But you might know better.

eregon:
Maybe we should have a Redmine ticket to discuss if strscan and digest might be removed from stdlib one day, or if there is no plan for that and therefore we should only need to be explicit when it will be needed?

nevans:
I don't know about _general_ rules.  My gut instinct is that we _should_ be moving toward explicitly specifying dependencies even on default gems.  But it's worthwhile looking at the _specific_ uses in `net-imap`:

`digest` and `strscan` are _only_ used by `CRAM-MD5` and `DIGEST-MD5` SASL authenticators, which have been deprecated for a decade or so.  No one should be using them.  We might keep them for "historical purposes", but I intend to submit a PR for #55, making them optional (they won't automatically be loaded by `require "net/imap"`).  Anyway, if there is a bug in `digest`, it's unlikely those mechanisms will be any better or worse after upgrading!  It _would_ be nice to add support for the `SCRAM-*` mechanisms, but they aren't widely deployed (as far as I can tell).  I doubt anyone will contribute that code any time soon.  If and when someone _does_ add support, I hope this whole situation will have improved somewhat?   The SASL authenticators which I _do_ intend to contribute in the very near future (`OAUTHBEARER` and `XOAUTH2`) do not require `digest`.

With respect to `strscan`, I've started constructing some sample IMAP response data to use in benchmarks of the `ResponseParser`.  I've started creating comparison lexers 1) using our current `case resp; when /regexp/` lexer and some minor variations, 2) using `StringScanner`, 3) generated by `kpeg`, and 4) generated by `ragel`.  _If_ the `StringScanner` approach wins, then I think I can easily promise it can remain backwards compatible with the current `StringScanner`... at least until this whole situation has a better resolution.

In addition to @hsbt's remark about `net-protocol`: as far as I can tell, our only dependency on `Net::Protocol` is `ssl_socket_connect`. Wouldn't a much better home for that code be in `OpenSSL::SSL::SocketForwarder#connect` anyway? 🙂 

Speaking of `openssl`: `net-imap` doesn't require it because it can connect to insecure servers without it.  However, in light of RFC8314, perhaps we *should* add a dependency to `openssl` in the future.  😉

nevans:
*shorter version:* I'm okay with removing all three dependencies, at least for now.

nevans:
@eregon In #58, I convert the digest and strscan dependencies to development dependencies.  That should remove the end-user pain, even if it doesn't _entirely_ solve the issue for truffleruby, jruby, etc.  If these are going to be gems at all, this problem isn't going to go away.  Maybe it would be worthwhile to update these gems to install on truffleruby and jruby with a no-op stub?

hsbt:
@nevans #58 seems good strategy for me. I agreed to reduce the dependencies in `net-imap`.

eregon:
Yes, that sounds helpful :)
I agree longer term TruffleRuby (& JRuby) need to support all those default gems.
But for TruffleRuby we'd like some time to consider what's the best way for that, and this helps in the meantime + unbreaks existing TruffleRuby releases for users for net-* gems.

headius:
> Maybe it would be worthwhile to update these gems to install on truffleruby and jruby with a no-op stub?

`digest` already supports JRuby, after some work last fall to move our implementation into the gem.

`strscan` has an open PR with full JRuby support.

We have been working on these default gems for some time now and intend to support all of the ones that have a JRuby equivalent.

PR #57, opened 2022-02-21
Adding RFC licenses

nevans:
I'm not entirely certain this is necessary (does our use of the RFCs for
documentation count as "fair use"?).  But it's an easy thing to add.


hsbt:
👍

PR #58, opened 2022-02-22
Do not load deprecated SASL mechanisms by default

nevans:
n.b. the mechanisms haven't been removed.  They just aren't loaded by
default.  Closes GH-55.

By making these optional, there's no reason to require the `digest` or
`strscan` gems anymore. Closes GH-56.

### DIFF
--- a/net-imap.gemspec
+++ b/net-imap.gemspec
@@ -32,6 +32,6 @@ Gem::Specification.new do |spec|
   spec.require_paths = ["lib"]
 
   spec.add_dependency "net-protocol"
-  spec.add_dependency "digest"
-  spec.add_dependency "strscan"
+  spec.add_development_dependency "digest"
+  spec.add_development_dependency "strscan"
 end
--- a/lib/net/imap.rb
+++ b/lib/net/imap.rb
@@ -380,26 +380,31 @@ module Net
     # the authentication mechanism to be used. Currently Net::IMAP
     # supports the authentication mechanisms:
     #
-    #   LOGIN:: login using cleartext user and password.
-    #   CRAM-MD5:: login with cleartext user and encrypted password
-    #              (see [RFC-2195] for a full description).  This
-    #              mechanism requires that the server have the user's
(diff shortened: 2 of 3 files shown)


nevans:
The list of non-deprecated SASL mechanisms is down to only `PLAIN`!  I'lt try to tidy up my `XOAUTH2` and `OAUTHBEARER` PR this week.

net-imap.gemspec
```diff
@@ -32,6 +32,6 @@ Gem::Specification.new do |spec|
   spec.require_paths = ["lib"]
 
   spec.add_dependency "net-protocol"
-  spec.add_dependency "digest"
-  spec.add_dependency "strscan"
+  spec.add_development_dependency "digest"
+  spec.add_development_dependency "strscan"
```
eregon, line 36: I'm not entirely how gem dev deps are treated by Bundler, but I guess they are ignored if e.g. a app Gemfile depends on `mail` or `net-imap`?

hsbt: They are ignored with `bundle install` by default.

hsbt (approved):

nevans:
:+1: I've been sitting on code for a couple of new SASL mechanisms, and a bugfix for "DIGEST-MD5" (when `frozen_string_literal: true` is being used).  I'll try and finish up the PRs for all of that today.  It would be a shame if the only "good" mechanism is "PLAIN"!  🙂

nevans:
I remembered why I hadn't pushed a PR for that other SASL code yet.  My refactorings went a little bit further than I wanted for a ticket like this, and I wasn't quite done yet!  But I might have that one ready for review soon, too.

At any rate, that branch reminded me that I wanted to try a different backward-compatible approach.  That approach and the DIGEST-MD5 bugfixes are here:
* #62

If you think that approach is okay, let's close this ticket and merge that one instead.

hsbt:
@nevans Sorry to late response. #62 is looks good. I'll close this.

Issue #52, opened 2021-12-28
Using PLAIN is broken in ruby 3.1.0


chdiza:
For years, up to and including ruby 2.7.5, I've been able to authenticate into some IMAP servers (even gmail) like so:
```
server = Net::IMAP.new(HOST, :port=>993, :ssl=>true)
server.authenticate('PLAIN', USERNAME, PASSWORD)
```
I've just upgraded to ruby 3.1.0 (from 2.7.5), and now this doesn't work on any of those servers.  Instead I get:
```
lib/ruby/gems/3.1.0/gems/net-imap-0.2.2/lib/net/imap.rb:1252:in `get_tagged_response': Authentication failed. (Net::IMAP::NoResponseError)
```

If I switch back to ruby 2.7.5 it works again.

I'm not knowledgeable enough to understand why this stopped working.

chdiza:
The problem also occurs in ruby 3.0.3, so I did some more digging.  Bisection reveals that this is the commit that broke authenticating with PLAIN:
```
23f241b081d94b6385e24f550a24e65a66a0d46d is the first bad commit
commit 23f241b081d94b6385e24f550a24e65a66a0d46d
Author: nicholas a. evans <nicholas.evans@gmail.com>
Date:   Tue Apr 27 16:33:27 2021 -0400

    Move each authenticator to its own file
    
    Also updates rdoc with SASL specifications and deprecations.  Of these
    four, only `PLAIN` isn't deprecated!
    
    +@@authenticators+ was changed to a class instance var
    +@authenticators+.  No one should have been using the class variable
    directly, so that should be fine.
```

chdiza:
Version 0.1.1 of the net-imap gem is fine.  But when I install a new ruby, `gem update` brings the buggy version of the gem into play.

shugo:
@chdiza Thanks for your report. I've just released net-imap-0.2.3.  Please try it.

chdiza:
Seems to work, thanks!!

nevans:
Yikes. Sorry about that. 😳 Thanks, @shugo!

PR #62, opened 2022-07-16
Warn when using deprecated SASL mechanisms

nevans:
Mark obolete SASL mechanisms as deprecated (fixes GH-55):
* This is a backwards-compatible alternative to the approach here: GH-58.  We can still use that incompatible approach in a future version.
* Warn every time a deprecated mechanism is used.
* Warnings can be disabled with `warn_deprecation: false`
* Fixes GH-56: delay loading standard gem dependencies until
  `#initialize`, and convert the gems to development dependencies.

Additionally:
* Adds basic tests for every authenticator (to avoid another GH-52!)
* Fixes a frozen string bug in DigestMD5Authenticator.
* Fixes constant resolution for exceptions in DigestMD5Authenticator.
* Can register an authenticator type that responds to #call (instead of
  #new).  I was originally going to register deprecated authenticators
  with a Proc that required the file and issued a warning, but I decided
  to put everything into the initializer instead.  `#authenticator`
  needed to be updated to safely delegate all args, and I left this in.

The DIGEST-MD5 bug was originally reported, tested, and fixed by
@singpolyma here: https://github.com/nevans/net-sasl/pull/3.

Co-authored-by: Stephen Paul Weber <singpolyma@singpolyma.net>

### DIFF
--- a/lib/net/imap.rb
+++ b/lib/net/imap.rb
@@ -378,28 +378,37 @@ module Net
     # Sends an AUTHENTICATE command to authenticate the client.
     # The +auth_type+ parameter is a string that represents
     # the authentication mechanism to be used. Currently Net::IMAP
-    # supports the authentication mechanisms:
+    # supports the following mechanisms:
     #
-    #   LOGIN:: login using cleartext user and password.
-    #   CRAM-MD5:: login with cleartext user and encrypted password
-    #              (see [RFC-2195] for a full description).  This
-    #              mechanism requires that the server have the user's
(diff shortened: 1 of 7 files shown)


shugo (approved):
looks fine. thank you!

hsbt:
@shugo @nevans Can we release `v0.3.0` contained this? I hope to reduce the dependencies from `gem i rails`.

hsbt:
Thanks a lot!

PR #65, opened 2022-10-14
Support `UIDPLUS` extension

### LINKED ISSUE #36: `RFC4315` (2005): `UIDPLUS` (included in `IMAP4rev2`)

hoffi:
Fixes #36.

Added the `UID EXPUNGE` command and parses the `APPENDUID`, `COPYUID` and `UIDNOTSTICKY` responses.
I also modified the `append`, `copy` and `move` methods to return those UIDs when they are present in the response.

I have also tested this against a production Dovecot IMAP server and a Microsoft 365 account.

### DIFF
--- a/lib/net/imap.rb
+++ b/lib/net/imap.rb
@@ -786,6 +786,15 @@ module Net
       end
     end
 
+    # Similar to #expunge, but takes a set of unique identifiers as
+    # argument.
+    def uid_expunge(set)
+      synchronize do
+        send_command("UID EXPUNGE", MessageSet.new(set))
+        return @responses.delete("EXPUNGE")
+      end
+    end
+
     # Sends a SEARCH command to search the mailbox for messages that
     # match the given searching criteria, and returns message sequence
     # numbers.  +keys+ can either be a string holding the entire
--- a/lib/net/imap.rb
+++ b/lib/net/imap.rb
@@ -435,8 +435,8 @@ module Net
     # in the +mailbox+ can be accessed.
     #
     # After you have selected a mailbox, you may retrieve the
-    # number of items in that mailbox from +@responses["EXISTS"][-1]+,
-    # and the number of recent messages from +@responses["RECENT"][-1]+.
+    # number of items in that mailbox from <code>@responses["EXISTS"][-1]</code>,
+    # and the number of recent messages from <code>@responses["RECENT"][-1]</code>.
     # Note that these values can change if new messages arrive
     # during a session; see #add_response_handler for a way of
     # detecting this event.
@@ -444,10 +444,11 @@ module Net
     # A Net::IMAP::NoResponseError is raised if the mailbox does not
     # exist or is for some reason non-selectable.
     #
-    # If the server supports the UIDPLUS extension it may return an additional
-    # "NO" response with a "UIDNOTSTICKY" response code indicating that the
-    # mailstore does not support persistent UIDs:
-    # +@responses["NO"].last.code.name == "UIDNOTSTICKY"+
+    # If the server supports the [UIDPLUS[https://www.rfc-editor.org/rfc/rfc4315.html]]
+    # extension it may return an additional "NO" response with a "UIDNOTSTICKY" response code
+    # indicating that the mailstore does not support persistent UIDs
+    # [1[https://www.rfc-editor.org/rfc/rfc4315.html#page-4]]:
+    #   @responses["NO"].last.code.name == "UIDNOTSTICKY"
     def select(mailbox)
       synchronize do
         @responses.clear
@@ -758,8 +759,9 @@ module Net
     # not exist (it is not created automatically), or if the flags,
     # date_time, or message arguments contain errors.
     #
-    # If the server supports the UIDPLUS extension it returns an array
-    # with the UIDVALIDITY and the assigned UID of the appended message.
+    # If the server supports the [UIDPLUS[https://www.rfc-editor.org/rfc/rfc4315.html]]
+    # extension it returns an array with the UIDVALIDITY and the assigned UID of the
+    # appended message.
     def append(mailbox, message, flags = nil, date_time = nil)
       args = []
       if flags
@@ -800,10 +802,27 @@ module Net
     end
 
     # Similar to #expunge, but takes a set of unique identifiers as
-    # argument.
-    def uid_expunge(set)
+    # argument. Sends a UID EXPUNGE command to permanently remove all
+    # messages that have both the \\Deleted flag set and a UID that is
+    # included in +uid_set+.
+    #
+    # By using UID EXPUNGE instead of EXPUNGE when resynchronizing with
+    # the server, the client can ensure that it does not inadvertantly
+    # remove any messages that have been marked as \\Deleted by other
+    # clients between the time that the client was last connected and
+    # the time the client resynchronizes.
+    #
+    # Note:: Although the command takes a +uid_set+ for its argument, the
+    #        server still returns regular EXPUNGE responses, which contain
+    #        a <em>sequence number</em>. These will be deleted from
+    #        #responses and this method returns them as an array of
+    #        <em>sequence number</em> integers.
+    #
+    # ==== Required capability
+    # +UIDPLUS+ - described in [UIDPLUS[https://www.rfc-editor.org/rfc/rfc4315.html]].
+    def uid_expunge(uid_set)
       synchronize do
-        send_command("UID EXPUNGE", MessageSet.new(set))
+        send_command("UID EXPUNGE", MessageSet.new(uid_set))
         return @responses.delete("EXPUNGE")
       end
     end
@@ -929,9 +948,9 @@ module Net
     # a number, an array of numbers, or a Range object. The number is
     # a message sequence number.
     #
-    # If the server supports the UIDPLUS extension it returns an array with
-    # the UIDVALIDITY, the UID set of the source messages and the assigned
-    # UID set of the copied messages.
+    # If the server supports the  [UIDPLUS[https://www.rfc-editor.org/rfc/rfc4315.html]]
+    # extension it returns an array with the UIDVALIDITY, the UID set of the source messages
+    # and the assigned UID set of the copied messages.
     def copy(set, mailbox)
       copy_internal("COPY", set, mailbox)
     end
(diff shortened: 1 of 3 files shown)


lib/net/imap.rb
```diff
@@ -443,6 +443,11 @@ def login(user, password)
     #
     # A Net::IMAP::NoResponseError is raised if the mailbox does not
     # exist or is for some reason non-selectable.
+    #
+    # If the server supports the UIDPLUS extension it may return an additional
+    # "NO" response with a "UIDNOTSTICKY" response code indicating that the
+    # mailstore does not support persistent UIDs:
+    # +@responses["NO"].last.code.name == "UIDNOTSTICKY"+
```
nevans, line 450: n.b. the simple rdoc `+` tends to not work over multiple words.  You need the `<code>` tag version for that.  More importantly, you'll get ruby syntax highlighting if you just indent it. 🙂

```suggestion
    #   @responses["NO"].last.code.name == "UIDNOTSTICKY"
```

lib/net/imap.rb
```diff
@@ -786,6 +799,15 @@ def expunge
       end
     end
 
+    # Similar to #expunge, but takes a set of unique identifiers as
+    # argument.
```
nevans, line 803: I have a documentation PR I've been working on (it's too big... I should just pick a stopping point and release it as is!).  So I'm going to make a suggestion here that uses the format I've been using in that branch. 😉

```suggestion
    # argument.  Sends a UID EXPUNGE command to permanently removes all
    # messages that have both the \Deleted flag set and a UID that is
    # included in arguments.
    # 
    # By using UID EXPUNGE instead of EXPUNGE when resynchronizing with
    # the server, the client can ensure that it does not inadvertantly
    # remove any messages that have been marked as \Deleted by other
    # clients between the time that the client was last connected and
    # the time the client resynchronizes.
    #
    # Note:: Although the command takes a +UID+ set for its argument, the
    #        server still returns regular EXPUNGE responses, which contain
    #        <em>sequence number</em>.  These will be deleted from
    #        #responses and this method returns them as an array of
    #        <em>sequence number</em> integers.
    #
    # ==== Required capability
    # +UIDPLUS+ - described in [UIDPLUS[https://www.rfc-editor.org/rfc/rfc4315.html]].
    #
    #--
    # TODO: when IMAP4rev2 is closer to being supported, add:
    #
    # or [IMAP4rev2[https://tools.ietf.org/html/rfc9051]], if it is enabled[#enable].
    #++
```

lib/net/imap.rb
```diff
@@ -906,6 +928,10 @@ def uid_store(set, attr, flags)
     # of the specified destination +mailbox+. The +set+ parameter is
     # a number, an array of numbers, or a Range object. The number is
     # a message sequence number.
+    #
+    # If the server supports the UIDPLUS extension it returns an array with
```
nevans, line 932: Adding a reference link.  🙂

```suggestion
    # If the server supports the [UIDPLUS[https://www.rfc-editor.org/rfc/rfc4315.html]]
    # extension, it returns an array with
```

n.b. this messes up the paragraph's line-breaks, but GitHub still doesn't allow multi-line suggestions.  😉

lib/net/imap.rb
```diff
@@ -443,6 +443,11 @@ def login(user, password)
     #
     # A Net::IMAP::NoResponseError is raised if the mailbox does not
     # exist or is for some reason non-selectable.
+    #
+    # If the server supports the UIDPLUS extension it may return an additional
```
nevans, line 447: Adding a reference.  🙂
```suggestion
    # If the server supports the [UIDPLUS[https://www.rfc-editor.org/rfc/rfc4315.html]]
    # extension it may return an additional
```
n.b. this messes up the paragraph's line-breaks, but GitHub still doesn't allow multi-line suggestions. 😉

lib/net/imap.rb
```diff
@@ -921,6 +947,10 @@ def uid_copy(set, mailbox)
     # a message sequence number.
     #
     # The MOVE extension is described in [EXT-MOVE[https://tools.ietf.org/html/rfc6851]].
+    #
+    # If the server supports the UIDPLUS extension it returns an array with
```
nevans, line 951:
```suggestion
    # If the server supports the [UIDPLUS[https://www.rfc-editor.org/rfc/rfc4315.html]]
    # extension it returns an array with
```

lib/net/imap/response_parser.rb
```diff
@@ -1321,6 +1341,23 @@ def number
         return token.value.to_i
       end
 
+      def set
```
nevans, line 1344: Can we rename this `uid_set` to more closely match the RFC4315 and RFC9051 names?  In addition to simply sticking as closely to the RFC grammar and names as we can, `sequence-set` and `uid_set` do have slightly different definitions.

```suggestion
      # RFC-4315 (UIDPLUS) or RFC9051 (IMAP4rev2):
      #      uid-set         = (uniqueid / uid-range) *("," uid-set)
      #      uid-range       = (uniqueid ":" uniqueid)
      #                          ; two uniqueid values and all values
      #                          ; between these two regardless of order.
      #                          ; Example: 2:4 and 4:2 are equivalent.
      #      uniqueid        = nz-number
      #                          ; Strictly ascending
      def uid_set
```

lib/net/imap/response_parser.rb
```diff
@@ -1119,6 +1125,20 @@ def resp_text_code
...
+          match(T_SPACE)
+          uidvalidity = number
+          match(T_SPACE)
+          append_uid = number
+          result = ResponseCode.new(name, [uidvalidity, append_uid])
+        when /\A(?:COPYUID)\z/n
+          match(T_SPACE)
+          uidvalidity = number
+          match(T_SPACE)
+          from_uid = set
```
nevans, line 1138: Can we rename this (see my comment on `def set` below)?

```suggestion
          from_uid = uid_set
```

lib/net/imap/response_parser.rb
```diff
@@ -1119,6 +1125,20 @@ def resp_text_code
...
+          match(T_SPACE)
+          to_uid = set
```
nevans, line 1140: Can we rename this (see my comment on `def set` below)?
```suggestion
          to_uid = uid_set
```

test/net/imap/test_imap.rb
```diff
@@ -775,6 +775,100 @@ def test_id
...
+        sock.close
+        server.close
+      end
+    end
+
+    begin
+      imap = Net::IMAP.new(server_addr, :port => port)
+      response = imap.uid_expunge(1000..1003)
+      assert_equal("RUBY0001 UID EXPUNGE 1000:1003\r\n", requests.pop)
+      assert_equal(response.length, 3)
```
nevans, line 804: What do you think about using this assertion instead?

```suggestion
      assert_equal(response, [1, 1, 1])
```

nevans (approved):
I've been using a regexp hack for almost a decade.  This is great, thanks!

lib/net/imap.rb
```diff
@@ -786,6 +799,15 @@ def expunge
       end
     end
 
+    # Similar to #expunge, but takes a set of unique identifiers as
+    # argument.
+    def uid_expunge(set)
```
nevans, line 804: No multi-line suggestions, yet!  😉

```suggestion
    def uid_expunge(uid_set)
```

lib/net/imap.rb
```diff
@@ -786,6 +799,15 @@ def expunge
...
+      synchronize do
+        send_command("UID EXPUNGE", MessageSet.new(set))
```
nevans, line 806: No multi-line suggestions, yet! 😉
```suggestion
        send_command("UID EXPUNGE", MessageSet.new(uid_set))
```

nevans (commented):
another tiny suggestion, a small rdoc improvement.

lib/net/imap.rb
hoffi: Ah yes, actually looked up the generated rdoc and also found some problems in the existing comment and fixed that as well.

lib/net/imap.rb
hoffi: Thanks! 😄
Something seems to be missing in the TODO comment. Not sure what is meant to be there?

lib/net/imap/response_parser.rb
hoffi: Ah yes that makes more sense :)

lib/net/imap.rb
nevans: Oh, sorry for the confusion.  I actually _did_ intend it (more-or-less) as is.  The `--` and `++` make those lines invisible for now.  The idea is to delete the `--` and `++` and `TODO` lines, and then the final version will look something like:

> `UIDPLUS` - described in [UIDPLUS].
>
> or [IMAP4rev2], if it is enabled.

But, reading it now, I want it changed again!  😉

The earlier (and mostly the current) style in this library was to follow the RFC style for citations, which are always enclosed in square brackets and (at least for the IMAP RFCs) prefer to use the extension name rather than the RFC number.

I think that makes sense for our documentation style when the capability or extension is mentioned elsewhere in the documentation.  But when the capability name is used explicitly as a name---like here---then "UIDPLUS [UIDPLUS]" isn't very helpful.  It reads much better to use the RFC number inside the citation: "UIDPLUS [RFC4315]".  Although that leads to inconsistent citations, I think that local readability is more important than global consistency.

Once we have full IMAP4rev2 support, it might look like one of the following two options:

A terse one line version: "`QUOTA` [RFC2087] or `IMAP4rev2` [RFC-9051]."
```ruby
    # ==== Required capability
    #
    # +QUOTA+ [RFC-2087[https://tools.ietf.org/html/rfc2087]]
    # or +IMAP4rev2+ [RFC-9051[https://www.rfc-editor.org/rfc/rfc9051.html]]
```

Or a longer two-line version:
> Either `QUOTA` [RFC2087] must be supported,
>
> or `IMAP4rev2` [RFC-9051] must be supported and enabled"
```ruby
    # ==== Required capability
    #
    # Either +QUOTA+ [RFC-2087[https://tools.ietf.org/html/rfc2087]] must be
    # supported,
    #
    # or +IMAP4rev2+ [RFC-9051[https://www.rfc-editor.org/rfc/rfc9051.html#section-6.4.4]]
    # must be supported and enabled[#enable].
```

What do you think?  I'm leaning towards longer, if only for the fact that it lets me link to #enable. (PR for #enable coming soon, maybe next week)

nevans: Anyway, don't worry about these (unless you want to! 😉).  What you have now is great and I'm merging it as-is.  I'll fold these documentation tweaks into my rdoc PR. (coming soon??!? 😉)

nevans:
🎉 Thanks!

lib/net/imap.rb
hoffi: Ahhh okay, yes that confused me 😄

I do like to have the RFC number in the citations as well as the longer version.
I find it more readable 🙂

PR #66, opened 2022-10-14
Fixes date-time format, and adds decode_datetime

nevans:
Fixes date-time format, and adds decode_datetime.

Also:
* adds "date" gem as a dependency
* adds Date handling to `send_data`.  This will be used by search.
* reduce duplication by using encode_datetime in send_time_data.

I don't know if anyone is actually *using* Net::IMAP.format_datetime. *If* they are, then fixing the bug will not be backwards compatible. Since "encode" and "decode" are a natural pair (and we already had "encode_utf7"), I added new methods with those names and added deprecation warnings to the old methods.

### DIFF
--- a/lib/net/imap/command_data.rb
+++ b/lib/net/imap/command_data.rb
@@ -1,5 +1,7 @@
 # frozen_string_literal: true
 
+require "date"
+
 require_relative "errors"
 
 module Net
@@ -21,7 +23,7 @@ module Net
             validate_data(i)
           end
         end
-      when Time
+      when Time, Date, DateTime
       when Symbol
       else
         data.validate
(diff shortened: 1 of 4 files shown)


nevans:
@hsbt @shugo My primary concern here is adding the "date" gem as a dependency.  I think it's the right thing to do.  It's been a default gem since 2.5 and is still compatible down to 2.4.  So there really shouldn't be any problems with it.

_...but..._

maybe there is?

nevans:
Looking at https://rubygems.org/gems/date/reverse_dependencies, it looks like there aren't any very large/popular gems with a dependency on `date`.  This would pull `date` into the `rails` dependency network.  I'm of the opinion that explicit stdlib gem dependencies are good, but they come with "growing pains" that we eventually need to just push through and fix.

@hsbt, @shugo, what do you think?

shugo (approved):
It's OK to add date.gem as a dependency.

mperham:
There is a major issue with `date`: it requires native compilation. My server does not have a compiler installed for security reasons and I cannot install the latest version of `net-imap` now.

I install Ruby from distro packages and then a set of pure Ruby gems but net-imap is no longer pure. ☹️

PR #67, opened 2022-10-14
Fixes "bundle exec rake" clash with test/unit

nevans:
When multiple versions of "test/unit" are installed, the Gemfile.lock can be out of sync with "test/unit" version selected without bundler. When that happens, "bundle exec rake" fails, even when "rake" passes.

This can generally be fixed by simply running "bundle update" so the lockfile has the latest version, too.  And the lockfile is gitignored so this doesn't affect CI.  But, the error can be confusing and waste time.

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -3,7 +3,7 @@ require "rake/testtask"
 
 Rake::TestTask.new(:test) do |t|
   t.libs << "test/lib"
-  t.ruby_opts << "-rhelper"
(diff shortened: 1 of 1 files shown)


nevans:
@hsbt I would have put `require "bundler/setup"` into `test/lib/helper.rb`, but I thought it was synchronized with `rake sync_tool`.  But I see now that `rake sync_tool` doesn't touch that file.

PR #68, opened 2022-10-14
Use ruby2_keywords for authenticator forwarding

nevans:
Without this, authenticators need to accept keywords for compatibility with older ruby versions.  Ruby 2.6 will send an empty hash for `**properties`, and that can crash (or be interpreted as an optional argument) if the authenticator doesn't accept keyword arguments.

Actually... is that a bad thing?  Maybe the error will motivate people to upgrade to ruby 2.7+ or contribute their custom authenticators to net-imap? 😉

It's easy to stay compatible with `ruby2_keywords`, so let's do that.

### DIFF
--- a/lib/net/imap.rb
+++ b/lib/net/imap.rb
@@ -15,6 +15,7 @@
 
 require "socket"
 require "monitor"
+require "ruby2_keywords"
 require 'net/protocol'
 begin
   require "openssl"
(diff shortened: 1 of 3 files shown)


shugo:
I'd like to drop Ruby 2.6 support instead,  because it's EOL.
What do you think?

nevans:
> I'd like to drop Ruby 2.6 support instead, because it's EOL. What do you think?

I think I'm going to add some pattern matching to my parser comparisons and benchmarks right away.  And... how do you feel about numbered parameters? 😉

More seriously, I wouldn't want to drop support unless there is a motivating feature or incompatibility.  But that is exactly what this ticket is. 🙂  Let's bump the minor version when we drop support.

As someone who maintains a few old apps that aren't always upgraded before the ruby EOL, I have appreciated when libraries give me an extra few months before they drop support.

But it has been a few months already now, the kwparam situation makes supporting everything before 2.7 more annoying than usual, and the new features from 2.7 and 3.0 are *very* enticing to me.  Also, I want to experiment with both Fiber.scheduler and Ractor support (to replace the receiver thread).  If we had something ready by then, I'd be very happy to drop 2.7 support as soon as April or May.

shugo:
@nevans 
I don't want to drop old Ruby versions support as long as the code as-is works.
However, I don't want to add code for EOL versions of Ruby.

It's OK for me to drop Ruby 2.7 support after it will reach EOL.

PR #70, opened 2022-11-21
Nodocs and remove warning

nevans:
I propose that:
* We should deprecate `client_thread`.  This PR only marks it with `:nodoc:`.
* We should ark `RESPONSE_ERRORS` with `:nodoc:`.  We don't need to break it if people are using it, but it's an internal implementation detail, in my opinion.
* Don't need to document the `openssl extensions must be installed` scenario.  We can still try to _support_ that use-case, but "openssl" has been a default gem for a long time.

  Although some installations may still disable it, this is very rare and not important enough to document here any more.  Looking at `net-http` and `net-smtp`, neither of them still document the scenario where "openssl" is missing.  Although it carefully avoids an implicit dependency, "net-http" doesn't try to catch any LoadError when an SSL connection is explicitly used.

### DIFF
--- a/lib/net/imap.rb
+++ b/lib/net/imap.rb
@@ -1269,9 +1269,7 @@ module Net
     #
     # port::  Port number (default value is 143 for imap, or 993 for imaps)
     # ssl::   If +options[:ssl]+ is true, then an attempt will be made
-    #         to use SSL (now TLS) to connect to the server.  For this to work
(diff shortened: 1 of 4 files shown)


nevans:
@shugo: the reference to `nevans/rdoc` has been removed from this PR.

shugo (approved):
It looks fine!

PR #73, opened 2022-11-21
🐛 Fix mailbox attrs

nevans:
I don't think many people *use* the Mailbox Attr and flags constants, but I had incorrectly capitalized some of them, so they didn't all work anyway.  I intend to submit a future PR that *does* use these constants for some simple predicate methods, and that requires fixing them first.

Additionally, the first commit 66240008d (shared with several other PRs) fixes a long-outstanding issue with RDoc: section comments aren't rendered.  I don't think that the way that I've fixed it here is ultimately the Right Way to fix it, but I want to push our updated documentation to https://ruby.github.io/net-imap, and my patched `rdoc` will work very well for that.

Once my rdoc fork was able to render the section comments, I made some significant formatting changes so it all reads much more nicely now.


shugo (commented):
It looks fine except "nevans/rdoc" dependency.

nevans:
@shugo Thanks for catching my `nevans/rdoc`.  I've pushed this branch without that dependency and I'll update the other branches too.

PR #76, opened 2022-11-21
ResponseParser docs

nevans:
There's no reason not to include at least very basic user-visible docs for ResponseParser.  I've been using Net::IMAP::ResponseParser (both separate from and combined with the rest of the Net::IMAP client, both my forked version and the upstream version) for over a decade. 😄

Also, I want to consider adding some parser options, e.g. "#{provider} quirks mode" vs "strict mode" vs "support experimental draft rfc", and it would make the most sense to document options those here.

### DIFF
--- a/lib/net/imap/response_parser.rb
+++ b/lib/net/imap/response_parser.rb
@@ -5,7 +5,9 @@ require_relative "errors"
 module Net
   class IMAP < Protocol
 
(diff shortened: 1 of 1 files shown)


nevans:
@shugo: the reference to `nevans/rdoc` has been removed from this PR.

shugo (approved):
It looks fine!

PR #75, opened 2022-11-21
Response Data docs

nevans:
I've converted the documentation for all of the response data structs' fields to meta-method rdoc.  (I prefer the rdoc formatting for methods vs attr.)  This allows us to link to the documentation for specific fields from elsewhere.

In several places, I did more than just copy the existing doecumentation to a new format.  E.g:
* Added relevant RFC links throughout.
* Added information about many standard codes to ResponseCode
* Added _much_ more detail was to FetchData, Envelope, Address, and BodyType*
* Added BodyStructure mixin module for the BodyType* structs
* Added deprecation warnings to BodyTypeExtension and BodyTypeAttachment.

This should be fairly complete, self-contained, and safe to merge as-is.

I have other changes in another branch (soon-to-be PR) that simplifies working with `Response => ResponseText => ResponseCode`, `BodyStructure` and the `BodyType*` classes, and makes the documented changes to `ResponseParser` so we can remove `BodyTypeExtension` and `BodyTypeAttachment`.  Those PRs will update these docs again, but I think it's fine to merge this first rather than wait for those PRs to be complete.

----

FYI: cherry-picks 66240008 (for rdoc improvements) and 76fef23e (to avoid conflicts, later)

PR #74, opened 2022-11-21
Net::IMAP Client docs

nevans:
This is a draft of my rdoc changes to the Net::IMAP class.  It cherry-picks from several other PRs:
* #70 
* #71 
* not just documentation, but it's needed because later documentation commits reference it:
  * #72

Some of the other commits are still incomplete.

PR #72, opened 2022-11-21
Add the UNSELECT command

nevans:
This PR fixes #40, which is one of the simplest missing commands for IMAP4rev2 support (#12).

PR #71, opened 2022-11-21
RFCs and references

nevans:
For simple ag/ack/rg/grep during development, I've added even more RFCs to the rake file.

Additionally, I updated and reformatted the "References" section... again.  See my comments on 981ea30.

PR #77, opened 2022-11-21
Many many documentation updates

nevans:
This is a meta-PR, combining/merging/cherry-picking documentation updates from:
* [x] #70
* [x] #71 - (but only the references updates, not the rake task update)
* [x] #72 - (because it's referenced by 06cdf6f and maybe another commit too)
* [x] #73 - (only the rdoc updates, not the bugfix)
* [ ] #74
* [x] #75 - (including the `BodyStructure` mixin)
* [x] #76

This branch provides a single stream of commits, without merge branching or conflicts.

I tried to keep the changes to only rdoc updates.  But a few other minor code changes were pulled in as well.

### DIFF
--- a/lib/net/imap.rb
+++ b/lib/net/imap.rb
@@ -70,6 +70,15 @@ module Net
   # UIDs have to be reassigned.  An \IMAP client thus cannot
   # rearrange message orders.
   #
+  # === Server capabilities and protocol extensions
+  #
+  # Net::IMAP <em>does not modify its behavior</em> according to server
+  # #capability.  Users of the class must check for required capabilities before
+  # issuing commands.  Special care should be taken to follow all #capability
+  # requirements for #starttls, #login, and #authenticate.
+  #
+  # See the #capability method for more information.
+  #
   # == Examples of Usage
   #
   # === List sender and subject of all recent messages in the default mailbox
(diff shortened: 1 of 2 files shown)


Gemfile
```diff
@@ -1,6 +1,10 @@
+# frozen_string_literal: true
+
 source "https://rubygems.org"
 
 gemspec
 
 gem "rake"
 gem "test-unit"
+
+gem "rdoc", github: "nevans/rdoc", branch: "darkfish-combine-call-seq-div"
```
shugo, line 10: I wouldn't add dependency outside ruby org.

Would you merge rdoc changes before merging this PR?

nevans: I agree.  That shouldn't be merged into master it is.  And that rdoc fork probably needs some more work before it makes a good PR.

I'd forgotten about this aspect when I based all of my other PRs on it.  I'll split the forked dependancy into another commit (for my own testing) and I'll post the rendered results somewhere.  But I'll remove that fork from all of these PRs.

Everything still works with the latest `rdoc` release, it just doesn't look as nice (IMO).  It's hard to read a class's Table of Contents when there's no nesting (h1-h4) and it's no good to use headings inside method docs unless there's more visual separation between each method.

nevans:
@shugo I made two PRs for ruby/rdoc: https://github.com/ruby/rdoc/pull/941 and https://github.com/ruby/rdoc/pull/942.

To demonstrate, I pushed a version with all of my rdoc updates here: https://nevans.github.io/net-imap/Net/IMAP.html.

PR #87, opened 2022-12-13
✅ Fix CI by pinning psych; Fixes #86

### LINKED ISSUE #86: Build broken by missing libyaml-dev
Psych doesn't vendor libyaml anymore, so the system needs the header files in order to install it.

See also:
* https://github.com/ruby/psych/issues/597
* https://github.com/ruby/setup-ruby/issues/409
* https://github.com/actions/runner-images/issues/6725

I'm hoping it's fixed in runner-images, because the conversation on `setup-ruby` was reluctant to fix it there.  IMO, setup-ruby should come with everything needed to install any standard gem without installing extra OS packages.

But, just to make the build work, we can probably pin to an older version of `psych` (temporarily).

nevans:
This is only needed temporarily, until actions/runner-images#6740 rolls out.

We *could* remove "rdoc" and use the system gem, but the "rake rdoc" customizations were written for 6.5.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -5,5 +5,10 @@ source "https://rubygems.org"
 gemspec
 
 gem "rake"
-gem "rdoc"
 gem "test-unit"
+
+# Some of the docs are difficult to read when rendered by earlier releases.
+# Customizations in rakelib/rdoc.rake were written for 6.5.
+gem "rdoc", ">=6.5"
+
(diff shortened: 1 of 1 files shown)


hsbt:
runner-images has been updated with `libyaml-dev`. We can use the latest version of `psych` now.

nevans:
🎉 great!  That was my preferred fix anyway.  :)

PR #89, opened 2022-12-21
⬆️ Drop v2.6 support; Require v2.7.3; Use "..." arg

nevans:
Argument forwarding was the issue in #68, and we decided to require ruby 2.7 instead of adding `ruby2_keywords`.

If I remember correctly, v2.7.0-v2.7.2 didn't allow `...` to be used with leading arguments, so this sets the minimum version to v2.7.3.


shugo (approved):
It looks fine!

PR #91, opened 2022-12-22
🔒 Add Net::IMAP#tls_verified?

nevans:
Returns true after the TLS negotiation has completed _and_ the remote hostname has been verified.

This can be used, for example, by automated safeguards against selecting particular SASL mechanisms—or against authenticating at all—when TLS hasn't been established _and_ verified the peer.


nevans:
missing tests...

nevans:
Added tests.

PR #92, opened 2022-12-22
✨ Add attr_readers for @host and @port

nevans:
My initial PR coerced host to a string and port to an Integer and froze both.  I removed all of that, so now the PR is simply `attr_reader :host, :port`.


nevans:
I converted this back to a draft because, upon reflection, I don't want to prevent using any host or port that `Addrinfo.getaddrinfo` can use.

shugo (approved):
It looks fine!

Issue #31, opened 2021-11-23
cache capabilities and add capability?(name)


nevans:
The client should keep track of server `CAPABILITY` responses and response codes.  And it should invalidate previously cached values at the appropriate times (e.g. after `STARTTLS`).

The base spec requires this: (e.g. #32 `LOGINDISABLED`, #48 `AUTH=`).  And many other extensions require the client to keep track (e.g. #37 `LITERAL-`). And the parser will also need to be able to "enable" different behavior for some extensions (e.g. 
`BINARY`, `UTF8=ACCEPT`, `IMAPrev2`).

PR #94, opened 2023-01-02
✨ Add greeting code data to responses

nevans:
This simplifies caching of the server capabilities, as many servers will send their capabilities inside the greeting.  It's needed for #31.

The PR is based on #93 only because my test for it is based on that branch.  But that PR is a bigger change than this one, so I can rebase this and give it its own test, if necessary.


shugo (approved):
It looks fine!

PR #97, opened 2023-01-07
🔊 Warn about deprecated `#responses` usage

nevans:
This was extracted from #93 and split into a separate PR.

This allows us to prepare dependent projects for the new behavior before their logs are flooded with deprecation warnings.

Also adds `Net::IMAP.attr_accessor :silence_thread_safety_deprecation_warnings`.  This is provided as a temporary workaround, until dependant projects can update their usage.

### DIFF
--- a/lib/net/imap/config.rb
+++ b/lib/net/imap/config.rb
@@ -306,7 +306,7 @@ module Net
         open_timeout: 30,
         idle_response_timeout: 5,
         sasl_ir: true,
-        responses_without_block: :silence_deprecation_warning,
+        responses_without_block: :warn,
       ).freeze
 
       @global = default.new


nevans:
@shugo Because #93 is such a big change to standard usage, I split the deprecation warnings into a separate PR.  I'd like to take a look at [the top reverse dependencies](https://rubygems.org/gems/net-imap/reverse_dependencies) and submit forward-compatible PRs there, before we start flooding their logs with new warnings!

shugo (dismissed):
It looks fine!

shugo:
FYI, `warn(msg, category: :deprecated)` doesn't show warnings by default.
However, Net::IMAP users may not use `-W:deprecated`....

nevans:
> FYI, `warn(msg, category: :deprecated)` doesn't show warnings by default. However, Net::IMAP users may not use `-W:deprecated`....

Thanks.  I had forgotten about that.

If I remember correctly, I did try to use `category: :deprecated` on the deprecated SASL mechanisms, but I removed it because that kwarg isn't supported in 2.7.3.  When we do require ruby 3.0+, should we switch to use `category: :deprecated`?

shugo:
> If I remember correctly, I did try to use category: :deprecated on the deprecated SASL mechanisms, but I removed it because that kwarg wasn't supported in 2.6 or 2.7. When we do require ruby 3.0+, should we switch to use category: :deprecated?

Yes, you're right.  I forgot that `category: :deprecated` was not supported in 2.7.

PR #98, opened 2023-01-13
Add support for ENABLE (RFC 5161)

arnt:
Hi,

this is almost a no-op because it doesn't change the parser.

Once this is merged, I'll contribute code to support UTF8=ACCEPT (RFC 6855), which depends on this, does extend the parser and has effect.

(There is a certain risk that I may implement QRESYNC as well.)

### DIFF
--- a/lib/net/imap.rb
+++ b/lib/net/imap.rb
@@ -199,8 +199,8 @@ module Net
   # [IDLE[https://tools.ietf.org/html/rfc2177]],
   # [NAMESPACE[https://tools.ietf.org/html/rfc2342]],
   # [UNSELECT[https://tools.ietf.org/html/rfc3691]],
+  # [ENABLE[https://tools.ietf.org/html/rfc5161]],
   #--
-  # TODO: [ENABLE[https://tools.ietf.org/html/rfc5161]],
   # TODO: [LIST-EXTENDED[https://tools.ietf.org/html/rfc5258]],
   # TODO: [LIST-STATUS[https://tools.ietf.org/html/rfc5819]],
   #++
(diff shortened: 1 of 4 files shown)


nevans:
Haha, coincidentally, I actually started to extract something like from a much larger branch of mine, just last night.  You may have saved me the effort of finishing that task.  😉

That other branch has _massive_ other changes to ResponseParser, and I was delaying posting it until I've finished adding test coverage and benchmarks.  I've fixed many many bugs, too... but I'm just not _quite_ done with the tests and benchmarks.  For what it's worth, that much larger branch _also_ has a parser that can handle CONDSTORE and QRESYNC.  But I might not have the time to be comfortable with merging it all of it into HEAD any time soon.  So I'll do some basic tidy-up and post what I have, before Monday.  You can take a look at it... and maybe take it for a test drive.  :)

nevans (approved):
Thanks again and sorry for the delay.  I've made some minor stylistic and documentation tweaks, and I did pull in some changes from my forthcoming parser updates.  Let me know what you think.

nevans:
Ahh, I rebased before merging from the command line, so GitHub didn't close the PR as merged. ☹️  I thought it would consider the rebased https://github.com/ruby/net-imap/commit/a40c0cfdeffa2300c8ab6213812d4f830b19e960 to be functionally identical to this PR (as the git CLI does).  Oops.

arnt:
Looks nice. Thanks.

You used double underscore in a couple of names, I assume that's a ruby convention with which I was not familiar. What does that mean?

nevans:
> You used double underscore in a couple of names, I assume that's a ruby convention with which I was not familiar. What does that mean?

It's not a ruby convention that I'm familiar with either. 😉  I'll add documentation to the code, but for now it's only documented in this commit: https://github.com/ruby/net-imap/commit/9caff31f2da100eb0dc32f5188897b4c55404901.

> Methods are all named according to the pattern:
>  ```ruby
>  [
>    modifier,
>    grammar_rule.tr("-", "_"),
>    subrule_or_modifier,
>  ].compact.join("__")
>  ```

I've been going through every rule in all of the IMAP RFCs' ABNF, documenting workarounds for common server bugs while fixing our own, etc. and this simplifies matching up the ABNF rules with the matching methods.  So, eg: `parens__object_id` = `"(" objectid ")"` and `nparens__objectid` = `"(" objectid ")" | nil` or rules like `mailbox-data` might be subdivided into sub-rules such as `mailbox_data__search` and `mailbox_data__status`, etc.

PR #101, opened 2023-01-21
✨ Add generic stringprep algorithm and the "trace" profile

nevans:
* Update `check_prohibited!`:
  * Fix error reporting: sending profile with the exception.
  * Update error reporting: `TABLE_REGEXPS.fetch` => `IndexError` with its "key not found: table" message is much more helpful than `nil.match?` => `NoMethodError`.
  * Accept regexp in addition to table names.
  * Add stringprep args for stored and unassigned: `stored: true` checks the `unassigned` table (with "A.1" as default).
* Add the RFC4422 mapping tables as hashes.
* Update rdoc for StringPrep tables

Defined in [RFC4505](https://tools.ietf.org/html/rfc4505) §3, The `trace` profile of StringPrep is used by the `ANONYMOUS` SASL mechanism.


nevans:
Now that `stringprep_trace` (and `saslprep`) in place, SASL `ANONYMOUS` (and `SCRAM-SHA-256`) are ready to ship next. 🙂

PR #103, opened 2023-02-04
✅ 📈 Move most parser tests to yaml, add more tests, and add parser benchmarks

nevans:
The two big motivations for this:
1) ✅ to simplify generation of test data from RFCs or other sources.
2) 📈 to simplify re-use of test data as benchmark data.

Currently defined test types:
* `:parser_assert_equal`: compares parse result to the `expected` data.
* `:parser_pending`: print out yaml which can be inserted into the
  appropriate yaml test fixture file.
* `:assert_parse_failure`: asserts `ResponseParseError` is raised

If there is no explicit test type, but there is a `:response` and an `:expected` key, then `:parser_assert_equal` is assumed.

If there is no explicit test type, but there is a `:response` without an `:expected` key, then `:parser_pending` is assumed.

This _should_ **simplify** adding new examples, but it needs to be documented.  We must not allow it to become a barrier for new contributors.

The benchmark script is very simply generated from all `:parser_assert_equal` tests.

----

TODO: document the simplest way to add new examples.

TODO: I have many more examples, but they need to be in separate PRs... because our current parser can't parse them yet.  😉

### DIFF
--- a/benchmarks/generate_parser_benchmarks
+++ b/benchmarks/generate_parser_benchmarks
@@ -1,0 +1,51 @@
+#!/usr/bin/env ruby
+
+require "yaml"
+require "pathname"
+require "net/imap"
+
+path = Pathname.new(__dir__) / "../test/net/imap/fixtures/response_parser"
+files = path.glob("*.yml")
+tests = files.flat_map {|file|
+  file.to_s
+    .then { YAML.unsafe_load_file _1 }
+    .fetch(:tests)
+    .select {|test_name, test|
+      :parser_assert_equal == test.fetch(:test_type) {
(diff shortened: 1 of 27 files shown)


nevans:
@shugo this is one half of why I've been holding back on any big parser updates.  The other half is that I want to test a few external dependent projects as well.  What do you think of this general direction?

nevans:
FYI, I added the following comment to partially document how I've been creating parser tests.  I can fill it out in more detail later, or add a script to simplify the creation of new parser tests even further.  But I think this is enough to merge the PR.

https://github.com/ruby/net-imap/blob/26547e6eca0e3751e31d73a3a2be294ee6ed56c3/test/net/imap/test_imap_response_parser.rb#L26-L29

Issue #49, opened 2021-11-23
Honoring server-reported and client-enabled capabilities


nevans:
Depends on:
- [ ] #31
- [ ] #33

Subtasks:
- [ ] Security related (authentication, TLS, etc)
  - [ ] #32
  - [ ] #48
  - [ ] Honor `STARTTLS` capability
- [ ] etc ...

### Using capabilities information

The client should track whenever it is knowingly violating the server capabilities.  At the very least, this should include checking capabilities before sending specific commands or command arguments.

#### backwards compatibility

For capabilities that are related to security, backwards-compatibility should require opt-in with a warning, which can be disabled.  e.g.:
````ruby
client.force_capability!("AUTH=PLAIN")` # => [*previous_capabilities, "AUTH=PLAIN"]
# $stderr << 'Warning: ignoring server capabilities for "AUTH=PLAIN"'
````

For capabilities that are unrelated to security, backwards-compatibility can be opt-out via a configuration option, e.g.:
````ruby
# Restores current behavior
client.ignore_server_capabilities = true

# The new default behavior: prints a warning to stderr
client.ignore_server_capabilities = nil

# Explicitly use the new default behavior, even if that default changes in the future
client.ignore_server_capabilities = :warn

client.ignore_server_capabilities = false # future default: raises an exception

# Allow client-configurable handling.
# Can be used to support capabilities that haven't been added to `net-imap`.
# The argument will be an CapabilityError object Which can be thrown.
# There should be at least one sub-class, SecurityCapabilityError, to represent "LOGINDISABLED", "AUTH=", "STARTTLS" etc.
client.ignore_server_capabilities = -> (error) {
  case error
  in AuthCapabilityError{sasl_mechanism: "FOOBAR"}
    raise "Server doesn't support our pretend SASL mechanism"
  in SecurityCapabilityError{attempted_command: "AUTH" | "STARTTLS"}
    raise SecureAuthUnsupportedError
  in SecurityCapabilityError
    raise error
  in attempted_command: "SELECT" | "EXAMINE", missing_capability: "QRESYNC"
    error.run_alternate_commands do
      # a theoretical API for gracefully degrading
    end
  in missing_capability: "X_PLEASE_WARN"
    logger.warn { "X_WARNING: error.warning" }
  in missing_capability: "X_PLEASE_WARN_2"
    # let Net::IMAP handle warning the user via $stderr
    :warn
  in missing_capability: "BINARY"
[...]

Issue #44, opened 2021-11-23
RFC4731     (2006):     `ESEARCH`               (included in `IMAP4rev2`)


nevans:
https://www.rfc-editor.org/rfc/rfc4731.html

nevans:
Fixed by #333, and released in [v0.5.2](https://github.com/ruby/net-imap/releases/tag/v0.5.2).

PR #114, opened 2023-02-22
Minor bits for UTF8=ACCEPT 

arnt:
This really just makes testing a little bit more convenient, and avoids a warning when someone uses a utf8 string in code like `foo.select("æ")`.

### DIFF
--- a/lib/net/imap.rb
+++ b/lib/net/imap.rb
@@ -1975,7 +1975,7 @@ module Net
       synchronize do
         send_command("ENABLE #{capabilities}")
         result = @responses.delete("ENABLED")[-1]
-        @utf8_strings ||= result.include? "SMTPUTF8"
+        @utf8_strings ||= result.include? "UTF8=ACCEPT"
         @utf8_strings ||= result.include? "IMAP4REV2"
         result
       end
--- a/lib/net/imap.rb
+++ b/lib/net/imap.rb
@@ -491,7 +491,7 @@ module Net
   #
   # ==== RFC6855: <tt>UTF8=ACCEPT</tt>, <tt>UTF8=ONLY</tt>
   #
-  # - See #enable for information about support foi UTF-8 string encoding.
+  # - See #enable for information about support for UTF-8 string encoding.
   #
   #--
   # ==== RFC7888: <tt>LITERAL+</tt>, +LITERAL-+
@@ -1974,7 +1974,10 @@ module Net
         .join(' ')
       synchronize do
         send_command("ENABLE #{capabilities}")
-        return @responses.delete("ENABLED")[-1]
+        result = @responses.delete("ENABLED")[-1]
+        @utf8_strings ||= result.include? "SMTPUTF8"
+        @utf8_strings ||= result.include? "IMAP4REV2"
+        result
       end
     end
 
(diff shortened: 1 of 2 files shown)


nevans (approved):
Thanks!  Could you change the rdoc for `UTF8=ACCEPT` at lines 1934-1937?   I think it would be fine to simply delete that paragraph.

nevans (changes requested):
Oops, I accidentally approved when I intended to "request changes".  The `SMTPUTF8` vs `UTF8=ACCEPT` needs to be resolved, at least.  🙂

arnt:
That's what I get for merging conflicts manual while jetlagged.

I adjusted the documentation; even with UTF8=ACCEPT it's perfectly legal to send a message with Content-Type: text/plain; charset=ebcdic, and retrieving something like BODY[1] will then retrieve EBCDIC. With BINARY (RFC 3510) you might also get a .zip file in a literal, without any encoding.

nevans:
This is great, thanks!

arnt:
One final comment.

I reviewed 9051 and 6855, and there are two differences. Both support UTF8 strings etc., but two things differ:

1. 6855 forbids `UID SEARCH CHARSET UTF8 SUBJECT rené`, while 9051 allows that. Ruby code will generally not send that, so I don't think there is a problem, but it's something you should be aware of.
2. 6855 offers extra `APPEND` syntax, which 9051 does not. net-imap does not use that, which is IMO good. All the syntax offers is a way to say "if there's any 8-bit in the headers of this message I'm appending, that's UTF8" so you can escape from 2047 hell. That's not a very generous offer.

AIUI you have a general plan of only sending the commands that are legal under both regimes, and I like that plan.

The other final comment is — ask me any favour. My favourite dynamic language plus these quick responses. What a combination. I'm charmed.

nevans:
Thanks again.  Your expertise is much appreciated!  I'll push a very simple documentation PR with your notes in it soon, maybe tonight or tomorrow.

As it currently stands, `net-imap` does little to no validation or normalization of most command arguments.  It has some basic encoding heuristics (e.g. arrays become parenthesized lists) but most of the time it is very "free form" and allows the users of the library to do what they will, for better and for worse.  This does make it very easy to "support" certain specifications: out of scope == not our problem.  😉  On the other hand, we should _at least_ provide documentation (with links into RFCs) for important points like these.

----

I do have plans to add more "smarts" to `net-imap`, to make a more "ergonomic" interface and to protect against basic, easily caught mistakes.  But I also want to preserve backward compatibility and retain the freedom to easily "cheat" a bit.  I'm slowly working on a set of "generic extensions" PRs right now, mostly for #35, #43, #44, and #49.  And they should simultaneously move us in the direction of more freedom and extensibility but also better definitions and validations, ultimately making it easier for users of the library to follow the specs.

To your first point, we currently require the user to provide the appropriate `charset`, sending nothing when `charset` is `nil`.  I intend to make `charset` an optional parameter and auto-detect the appropriate charset when none is explicitly given or raise an exception when incompatible charsets are used.

To your second point, I'm glad you don't consider the literal8/utf8-literal extensions to append to be a priority... because I wasn't prioritizing them either (I do have an implementation of BINARY FETCH nearly ready to go).  I'm assuming that nearly every ruby script or app that touches email uses the `mail` gem, and (IIRC) although it can parse UTF8 email, it still _generates_ US-ASCII by default.

nevans:
> we currently require the user to provide the appropriate `charset`

I don't know why I was thinking this.  It's already optional, as it should be.  I must have been looking at the private internal_search method.  (I still intend to add auto-detection code as part of my RFC4466 return opts commit.)

Anyway, I'm incorporating your notes into the rdoc for search, append, and enable, right now. 🙂

PR #135, opened 2023-03-09
📚 Auto-deploy GitHub Pages from an action

nevans:
Copied from ruby/rdoc#981 (thanks @hsbt!).  I modified it slightly, to set the `upload-pages-artifact` path rather than the `rdoc.path`.

This replaces the `rake ghpages` task from #102.

As far as I can tell, the only settings I needed to update on my own nevans/net-imap repository were the branch protection rules for the `github-pages` environment.  But it looks like the `github-pages` environment here already has the appropriate config.


hsbt:
👍

PR #137, opened 2023-03-10
✅ Add RFC3454 data, to support offline testing

### LINKED ISSUE #136: test suite cannot be run offline
Hi,

When attempting to run the test suite with `rake test`, it appears to attempt reaching to the network to fetch RFCs:

```
starting phase `check'
rake aborted!
SocketError: Failed to open TCP connection to www.rfc-editor.org:443 (getaddrinfo: Name or service not known)
/gnu/store/2y2hkxzygxqcxy4mzdfkw50j6s88jjw7-ruby-timeout-0.3.2/lib/ruby/vendor_ruby/gems/timeout-0.3.2/lib/timeout.rb:189:in `block in timeout'
/gnu/store/2y2hkxzygxqcxy4mzdfkw50j6s88jjw7-ruby-timeout-0.3.2/lib/ruby/vendor_ruby/gems/timeout-0.3.2/lib/timeout.rb:196:in `timeout'
/tmp/guix-build-ruby-net-imap-0.3.4.drv-0/source/rakelib/rfcs.rake:164:in `block in <top (required)>'

Caused by:
SocketError: getaddrinfo: Name or service not known
/gnu/store/2y2hkxzygxqcxy4mzdfkw50j6s88jjw7-ruby-timeout-0.3.2/lib/ruby/vendor_ruby/gems/timeout-0.3.2/lib/timeout.rb:189:in `block in timeout'
/gnu/store/2y2hkxzygxqcxy4mzdfkw50j6s88jjw7-ruby-timeout-0.3.2/lib/ruby/vendor_ruby/gems/timeout-0.3.2/lib/timeout.rb:196:in `timeout'
/tmp/guix-build-ruby-net-imap-0.3.4.drv-0/source/rakelib/rfcs.rake:164:in `block in <top (required)>'
Tasks: TOP => test => saslprep_rb => lib/net/imap/sasl/stringprep_tables.rb => rfcs/rfc3454-stringprep_tables.json => rfcs/rfc3454.txt
(See full trace by running task with --trace)
error: in phase 'check': uncaught exception:

```

It'd be nice if this was not needed and the test suite could be run in an isolated environment.

Thanks!

nevans:
We generate `rfcs/rfc3454-stringprep_tables.json` from RFC3454, and generate the regexps in both `lib/net/imap/stringprep/tables.rb` and `lib/net/imap/stringprep/saslprep_tables.rb` using those tables extracted from RFC3454.  The JSON data is also used to validate that the regexps are still valid, as we use Unicode character classes which can be updated every year, and thus with every new version of ruby.

However, this means that the test suite cannot be run offline (unless you have previously downloaded RFC3454).  Adding the text version of RFC3454 solves this issue.

Fixes #136.

### DIFF
--- a/net-imap.gemspec
+++ b/net-imap.gemspec
@@ -25,7 +25,7 @@ Gem::Specification.new do |spec|
   # Specify which files should be added to the gem when it is released.
   # The `git ls-files -z` loads the files in the RubyGem that have been added into git.
   spec.files         = Dir.chdir(File.expand_path('..', __FILE__)) do
-    `git ls-files -z 2>/dev/null`.split("\x0").reject { |f| f.match(%r{^(bin|test|spec|features)/}) }
+    `git ls-files -z 2>/dev/null`.split("\x0").reject { |f| f.match(%r{^(bin|test|spec|features|rfcs)/}) }
   end
   spec.bindir        = "exe"
   spec.executables   = spec.files.grep(%r{^exe/}) { |f| File.basename(f) }
--- a/rfcs/rfc3454.txt
+++ b/rfcs/rfc3454.txt
@@ -1,0 +1,5099 @@
+
+
+
+
+
+
+Network Working Group                                         P. Hoffman
+Request for Comments: 3454                                    IMC & VPNC
+Category: Standards Track                                    M. Blanchet
+                                                                Viagenie
+                                                           December 2002
+
+
+        Preparation of Internationalized Strings ("stringprep")
+
+Status of this Memo
+
+   This document specifies an Internet standards track protocol for the
+   Internet community, and requests discussion and suggestions for
+   improvements.  Please refer to the current edition of the "Internet
+   Official Protocol Standards" (STD 1) for the standardization state
+   and status of this protocol.  Distribution of this memo is unlimited.
+
+Copyright Notice
+
+   Copyright (C) The Internet Society (2002).  All Rights Reserved.
+
+Abstract
+
+   This document describes a framework for preparing Unicode text
+   strings in order to increase the likelihood that string input and
+   string comparison work in ways that make sense for typical users
+   throughout the world.  The stringprep protocol is useful for protocol
+   identifier values, company and personal names, internationalized
+   domain names, and other text strings.
+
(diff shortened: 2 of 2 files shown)


nevans:
@hsbt FYI: Because this is only needed to allow self-contained offline testing and isn't ever used at run-time, I'm excluding the `rfcs` dir from the gemspec files.

Apteryks:
Hello!

$ ./pre-inst-env guix build ruby-net-imap --with-branch=ruby-net-imap=offline-test-suite

[...]
starting phase `check'
Loaded suite /gnu/store/j4z07lyi1ykk8bc68h1p4bpj1il9dn3f-ruby-2.7.4/lib/ruby/gems/2.7.0/gems/rake-13.0.1/lib/rake/rake_test_loader
Started
...............................................................................
...............................................................................
..........................
Finished in 14.324993697 seconds.
-------------------------------------------------------------------------------
184 tests, 362 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed
-------------------------------------------------------------------------------
12.84 tests/s, 25.27 assertions/s
phase `check' succeeded after 16.6 seconds
[...]
```
LGTM!

Thanks for the timely fix!

hsbt (commented):
It's okay to me technically. But I'm not sure what's a license of distribution RFC text.

@nevans Can you do clearly this?

nevans:
> It's okay to me technically. But I'm not sure what's a license of distribution RFC text.

@hsbt: I was wondering that as well.  The copyright statement on RFC3454 seems to allow it: "provided that the above copyright notice and this paragraph are included on all such copies and derivative works".  Since the entire RFC txt file has been included, unmodified, the copyright notice is included as well:
https://github.com/ruby/net-imap/blob/51e08fbd7a223f2a4697f5a88ecdffdcc662590d/rfcs/rfc3454.txt#L5047-L5073

(Alternatively, we could add the `rfcs/rfc3454-stringprep_tables.json` file and update the `rake test` dependencies so it doesn't attempt to regenerate itself.  But adding the RFC seems like the simplest approach to me.)

PR #152, opened 2023-06-09
Decode UTF-7 more strictly

nobu:
Reported by svalkanov in <https://hackerone.com/reports/1969040>.

### DIFF
--- a/lib/net/imap/data_encoding.rb
+++ b/lib/net/imap/data_encoding.rb
@@ -54,9 +54,9 @@ module Net
     # Net::IMAP does _not_ automatically encode and decode
     # mailbox names to and from UTF-7.
     def self.decode_utf7(s)
-      return s.gsub(/&([^-]+)?-/n) {
-        if $1
-          ($1.tr(",", "/") + "===").unpack1("m").encode(Encoding::UTF_8, Encoding::UTF_16BE)
+      return s.gsub(/&([A-Za-z0-9+,]+)?-/n) {
+        if base64 = $1
+          (base64.tr(",", "/") + "===").unpack1("m").encode(Encoding::UTF_8, Encoding::UTF_16BE)
         else
           "&"
         end
--- a/test/net/imap/test_imap_data_encoding.rb
+++ b/test/net/imap/test_imap_data_encoding.rb
@@ -29,6 +29,10 @@ class IMAPDataEncodingTest < Test::Unit::TestCase
     s = Net::IMAP.decode_utf7("&,yH,Iv8j-")
     utf8 = "\357\274\241\357\274\242\357\274\243".dup.force_encoding("UTF-8")
     assert_equal(utf8, s)
+
+    assert_linear_performance([1, 10, 100], pre: ->(n) {'&'*(n*1_000)}) do |s|
+      Net::IMAP.decode_utf7(s)
+    end
   end
 
   def test_encode_date


nevans:
@shugo Can I sign up somewhere to get access to Hacker One reports for this project?

nevans:
@shugo @hsbt Because `master` has so many changes, including dropping support for ruby 2.6, and because this is a security fix, I'd like to propose that we release `v0.3.5` from the `v0.3-bugfixes` branch.

shugo:
> @shugo @hsbt Because `master` has so many changes and features, including dropping support for ruby 2.6, I had hoped to bump the next released version up to `v0.4.0`. And because this is a security fix, I'd like to propose that we release this as `v0.3.5` from the [v0.3-bugfixes](https://github.com/ruby/net-imap/tree/v0.3-bugfixes) branch.
> 
> I don't expect that we'll backport _every_ bugfix. Just the simple or very important ones.

We've decided to handle this as a non-security issue, so would you release just 0.4.0?

decode_utf7 is intended to use for mailbox names, so DoS happens only if an application connects to a malicious IMAP server.   It's unlikely, and if it happens, other bad things like credential leakage will happen without this issue.

We've sent a disclosure request and are waiting the reporter's approval.

shugo:
> @shugo Can I sign up somewhere to get access to Hacker One reports for this project? I assume it's connected to https://hackerone.com/ruby.

@hsbt Would you invite him?

shugo:
> > @shugo @hsbt Because `master` has so many changes and features, including dropping support for ruby 2.6, I had hoped to bump the next released version up to `v0.4.0`. And because this is a security fix, I'd like to propose that we release this as `v0.3.5` from the [v0.3-bugfixes](https://github.com/ruby/net-imap/tree/v0.3-bugfixes) branch.
> > I don't expect that we'll backport _every_ bugfix. Just the simple or very important ones.
> 
> We've decided to handle this as a non-security issue, so would you release just 0.4.0?

It's fine to release 0.3.5 if you would like to backport not only security fixes.

shugo:
@nevans hsbt invited you before.  Could you check your mailbox?

nevans:
@shugo haha, of course @hsbt did, and I missed it. Thanks!

I ran the provided benchmark and... WOW: I really need to upgrade all of my servers to ruby 3.2!  3.2 blasted through every length string at full speed.  But 3.1 and lower really struggled on longer malicious mailbox names.

nevans:
I agree with your assessment for the most common scenario of a user connecting to a malicious server.  For this reason, I'm okay with handling as non-security issue.

But there is a different scenario that I need to consider (because it affects systems that I'm responsible for): a service that connects to user-specified IMAP servers and can be forced to issue a `LIST` command.  A attacker could direct the service to connect to a malicious server under the attacker's control, and that could easily be leveraged for a DoS attack.  Because it could tie up the GVL for seconds, I _do_ consider it a vulnerability.

However, a very generous limit on mailbox length--maybe 1000 to 2000--should almost completely block this attack.  A service like this should have these limits _anyway_, unrelated to the ReDoS attack (but it would need to have implemented the limit _before_ decoding).  And other standard scaling techniques (all of which should be implemented anyway) can also significantly mitigate its impact.  Because of this, I consider it low severity.

Honestly, the bugs and bizarre unexpected behaviors of very popular non-malicious servers have combined with my own buggy or non-robust error handling to create many "DoS attacks" that are more severe than this.  😉  So, because it only affects the special scenario of services that connect to any user-specified IMAP server, I think it's fair to treat it as a non-security issue.

nevans:
Some examples of simple scaling techniques that would probably need to be implemented anyway, in order to scale a service like this, which should mitigate the impact of this and many other DoS attacks:
* most relevant to _this_ issue: length limit on mailbox name
* length limit on total response size
* queues with backpressure
* "fair" queues (so one user can't steal all of a server's time from other users)
* rate limiting new account formation
* rate limiting bandwidth
* using evented IO (e.g. with `event_machine` and/or fibers, or `async`)
* partitioning clients into multiple processes (thus multiple GVLs)
* timeouts
* exponential backoff when retrying failures

shugo:
Thanks for your confirmation.

I agree that such a length limit may be helpful.

nevans:
> We've decided to handle this as a non-security issue, so would you release just 0.4.0?
...
> It's fine to release 0.3.5 if you would like to backport not only security fixes.

@shugo FYI: I released 0.3.5 from the `v0.3-bugfixes` branch (also including a ruby 2.6 compatibility fix that I had to write for an older deployment).  There were a few minor issues I wanted to address before releasing 0.4.0, but I can release that by the end of the week too.

Thanks.

PR #174, opened 2023-09-19
🔒 Add `ssl_ctx` and `ssl_ctx_params` attr readers

nevans:
`Net::IMAP#ssl_ctx` will return the `OpenSSL::SSL::SSLContext` used by the `OpenSSL::SSL::SSLSocket` when TLS is attempted, even when the TLS handshake is unsuccessful.  The context object will be frozen.  It will return `nil` for a plaintext connection.

`Net::IMAP#ssl_ctx_params` will return the parameters that were sent to `OpenSSL::SSL::SSLContext#set_params` when the connection tries to use TLS (even when unsuccessful).  It will return `false` for a plaintext connection.

Additionally, error handling for the arguments to `#starttls` was moved *before* sending the command to the server.  This way a simple argument error will still raise an exception, but won't result in dropping the connection.  Because `ssl_ctx` is also created ahead of time, this will also catch invalid SSLContext params.

Tangentially:
* `#initialize` was moved to the top of the `public` section of the class definition, to match common conventions.
* Various other rdoc updates were made to `#initialize` and `#startls` rdoc, including examples, links to `OpenSSL::SSL::SSLContext#set_params`, etc.
* `#initialize` was refactored into several extracted private methods, for readability.

### DIFF
--- a/lib/net/imap.rb
+++ b/lib/net/imap.rb
@@ -722,34 +722,79 @@ module Net
     # The port this client connected to
     attr_reader :port
 
-    # :call-seq:
-    #    Net::IMAP.new(host, options = {})
-    #
     # Creates a new Net::IMAP object and connects it to the specified
     # +host+.
     #
-    # +options+ is an option hash, each key of which is a symbol.
+    # ==== Options
     #
-    # The available options are:
+    # Accepts the following options:
     #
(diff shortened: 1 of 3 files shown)


nevans:
_Note:_ This PR accidentally included rdoc references to `DeprecatedClientOptions`, which will be added by #175.

PR #175, opened 2023-09-19
🗑️ Deprecate backward compatible parameters to `new` and `starttls`

nevans:
Since 2007, `.new` and `#starttls` have taken an options hash (bad00272) and the original parameters have been undocumented (e631911e).  The original `usessl`, `certs`, and `verify` parameters will now print a deprecation warning whenever they are used.  In a future release (no sooner than a year after 0.4.0 is released), they will be removed from both methods.

The `options` hash has been converted to keyword parameters.  An. sending `port` as the second argument to `#initialize` is also obsolete.  Neither of them will print a deprecation warning, but they _will_ be marked as deprecated in a future release.

To simplify the implementations of both `#initialize` and `#starttls`, the backwards compatible parameters have all been extracted to a `DeprecatedClientOptions` module, which is (currently) prepended to `Net::IMAP`.  

_Note: #174 accidentally included rdoc references to `DeprecatedClientOptions`, which is added by this PR._

### DIFF
--- a/lib/net/imap.rb
+++ b/lib/net/imap.rb
@@ -1075,7 +1075,12 @@ module Net
     # Sends a {STARTTLS command [IMAP4rev1 §6.2.1]}[https://www.rfc-editor.org/rfc/rfc3501#section-6.2.1]
     # to start a TLS session.
     #
-    # Any +options+ are forwarded to OpenSSL::SSL::SSLContext#set_params.
+    # Any +options+ are forwarded directly to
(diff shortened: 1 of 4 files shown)


nevans:
FYI: the `mail` gem currently uses the deprecated options to `.new`:
https://github.com/mikel/mail/blob/b6b6cb737d47a85ddc720fda0e6b991e99224848/lib/mail/network/retriever_methods/imap.rb#L167
```ruby
        imap = Net::IMAP.new(settings[:address], settings[:port], settings[:enable_ssl], nil, false)
```

Not only that, but it sets `verify_mode: VERIFY_NONE` too!

PR #172, opened 2023-09-10
🔒 Add SASL SCRAM-SHA-* mechanisms

nevans:
Loosely based an initial implementation by @singpolyma at
https://github.com/nevans/net-sasl/pull/5

* Depends on #179
* Fixes #54

PR #178, opened 2023-09-22
✨ Add #logout! to combine logout and disconnect 

nevans:
A logout command that also disconnects, even if logout raises an exception, and doesn't raise exceptions when already disconnected would be very useful in situations where the connection MUST be gracefully closed, e.g. for security or tests.

Although the change is relatively insignificant, changing #logout to also ignore exceptions and immediately call #disconnect would be backwards incompatible.  So a new method was created.

### DIFF
--- a/lib/net/imap.rb
+++ b/lib/net/imap.rb
@@ -868,7 +868,7 @@ module Net
 
     # Disconnects from the server.
     #
-    # Related: #logout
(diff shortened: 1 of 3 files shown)


nevans:
I wrote this as part of #179 (which was itself written as part of #172).

PR #183, opened 2023-09-26
🔒⚗️ Add experimental SASL::ClientAdapter

nevans:
This adds some complexity over the IMAP implementation, but it simplifies sharing the implementation between `net-imap`, `net-smtp`, and hopefully many others in the future.

~`SASL.initial_response?` and `SASL.done?` have been removed without deprecation (they haven't been in any release yet).  The logic has been moved into `SASL::ClientAdapter`.~ Done in #184.

~`#initial_response?` was added to the deprecated authenticators.  It isn't required that authenticators implement this method, but it simplifies the tests, now that `SASL.initial_response?` is gone.~ Done in #184.

_This PR has been updated to only _create_ the SASL::ClientAdapter.  Actually using it has been pushed off until a later PR: #194._

### DIFF
--- a/lib/net/imap.rb
+++ b/lib/net/imap.rb
@@ -670,8 +670,9 @@ module Net
       "UTF8=ONLY" => "UTF8=ACCEPT",
     }.freeze
 
-    autoload :SASL,       File.expand_path("imap/sasl",       __dir__)
-    autoload :StringPrep, File.expand_path("imap/stringprep", __dir__)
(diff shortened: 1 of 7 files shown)


nevans:
This PR has been updated to only create the SASL::ClientAdapter. Actually using it has been pushed off until a later PR: #194.  As such, I see it as very low risk.

On the other hand, I have been _using_ this code from other projects and I would like to submit PRs to some of them.  And that will be greatly simplified once it is in a released gem.

PR #185, opened 2023-09-29
🐛 Fix typo in uncommon BODYSTRUCTURE parsing code

nevans:
I believe this code is only executed when there is something strange in the server's response.  If I remember correctly, I've usually only encountered this from server bugs... but it's an easy work around.


nevans:
The examples that I have in my current tests that trigger this bug are also broken in other ways, so I'll add another PR with those examples and the full fix (after v0.4.0).

PR #194, opened 2023-10-09
♻️ Use SASL::ClientAdapter

nevans:
Use the experimental adapter created by #183.


nevans:
I'd been delaying this because I wanted to handle cancelation properly.  But that isn't necessary: cancelation is no worse in the new implementation than it's ever been.  So I split that work off into #324.

PR #210, opened 2023-10-28
Set utf-8 encoding when looking for VERSION in the file.

debasishbsws:
Signed-off-by: Debasish Biswas <debasishbsws.dev@gmail.com>

By default, Ruby reads files with the system's default encoding, which is not utf-8 for every system and it does not handle special characters like '§' correctly and fails to get the VERSION from the lib/net/imap.rb file as it contains special characters
https://github.com/ruby/net-imap/blob/6dda58178b47d402edb0486e48d10c7558bc9b60/lib/net/imap.rb#L149-L151


nevans (approved):
Thanks!

Issue #115, opened 2023-02-23
RFC4422 2.6.1: Extended `SEARCH` command


nevans:

nevans:
Fixed by #333, and released in [v0.5.2](https://github.com/ruby/net-imap/releases/tag/v0.5.2).

PR #251, opened 2023-12-14
Remove redundant calls in sort_internal and thread_internal

gobijan:
sort_internal and thread_internal have an else statement that can be removed because normalize_searching_criteria(search_keys) is called anyways after the case.

### DIFF
--- a/lib/net/imap.rb
+++ b/lib/net/imap.rb
@@ -2880,8 +2880,9 @@ module Net
     def sort_internal(cmd, sort_keys, search_keys, charset)
       if search_keys.instance_of?(String)
         search_keys = [RawData.new(search_keys)]
+      else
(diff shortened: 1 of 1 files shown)


nevans:
Thanks!  I was looking at the `normalize_searching_criteria` code just last night, in order to finish up a PR for #115, and I was puzzling over why it might be called twice like that!  I hadn't realized there was a PR waiting to fix it.  😆  Maybe it was an accident during 080bd612?  It's lasted twenty years!

Could you update your PR to match the style `search_internal`?  We don't need to `normalize_searching_criteria` at all when the criteria is just `RawData`.

gobijan:
:) Done

nevans (approved):

PR #257, opened 2023-12-27
📚 Workaround rdoc method visibility bug

nevans:
Apparently rdoc doesn't recognize the visibility modifiers when they are used with method def arguments? 🙁


nevans:
It's probably unfair to call it an rdoc "bug"... but it is at least an unexpectedly missing feature.

nevans:
I must have messed up my testing of this somehow... I'm still seeing private methods in the SequenceSet docs. 😦

PR #259, opened 2024-01-01
Renew test certificates

sorah:
Replicating https://github.com/ruby/net-http/pull/169

----

The private key is replaced with a public known test key published at [RFC 9500].

Also lifetime has been extended to 10 years from 4 years.

[RFC 9500]: https://www.rfc-editor.org/rfc/rfc9500.html


nevans:
Thanks, @sorah!

PR #261, opened 2024-01-17
Add base64 dev dependency

hsbt:
`base64` is extracted as bundled gems for Ruby 3.4. We should add it to dev deps.

### DIFF
--- a/net-imap.gemspec
+++ b/net-imap.gemspec
@@ -35,8 +35,4 @@ Gem::Specification.new do |spec|
 
   spec.add_dependency "net-protocol"
   spec.add_dependency "date"
-
-  spec.add_development_dependency "digest"
-  spec.add_development_dependency "strscan"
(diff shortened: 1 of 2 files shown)


nevans (approved):

net-imap.gemspec
```diff
@@ -35,7 +35,4 @@ Gem::Specification.new do |spec|
 
   spec.add_dependency "net-protocol"
   spec.add_dependency "date"
-
-  spec.add_development_dependency "digest"
```
nevans, line 39: I'm curious: will development dependencies be deprecated in favor of Gemfile dependencies, or is there documentation on when they are still appropriate?  Thanks!

hsbt: `add_development_dependency` is still accepted entirely. I prefer to use Gemfile instead of it because Gemfile has many of useful feature.

PR #268, opened 2024-02-07
✅ Update Github test workflow name

nevans:
Tests are run on both Ubuntu and Mac OS, so the name "ubuntu" was misleading.  🙂


nevans:
@hsbt I'm tagging you in case you have scripts or other other automation that might be affected by changing the GitHub actions workflow name.

PR #271, opened 2024-04-19
Consider extra empty space in BODYSTRUCTURE

gaynetdinov:
Hi! I'm working on (very old) project that's excessively using net-imap to access remote IMAP servers.  Since ages we have some 'imap hacks' implemented on top of net-imap. These hacks were introduced to cope with some unusual responses from IMAP servers of our customers. I thought that I'd try to 'backport' some of them if you'd consider them viable/useful. 

In this my first PR I'm 'backporting' one fix we had since 2016. Responses from some IMAP servers contained extra space like so `((\"text\" \"plain\"  (\"charset\" \"UTF-8\")`. The fix seems to be easy and should not do harm I guess. I'm not sure if the fix is idiomatic according to the recent changes in net-imap (we're bumping net-imap frim 0.3.7 to 0.4.10 right now, so how we implemented the fix long time ago might not be pretty in 0.4.10).

Thank you for your time!

### DIFF
--- a/lib/net/imap/response_parser.rb
+++ b/lib/net/imap/response_parser.rb
@@ -1155,9 +1155,9 @@ module Net
       # RFC3501, RFC9051:
       # body-fld-param  = "(" string SP string *(SP string SP string) ")" / nil
       def body_fld_param
+        quirky_SP? # See comments on test_bodystructure_extra_space
         return if NIL?
         param = {}
-        shift_token if @token.symbol == T_SPACE
         lpar
         name = case_insensitive__string; SP!; param[name] = string
         while SP?
--- a/lib/net/imap/response_parser.rb
+++ b/lib/net/imap/response_parser.rb
@@ -1157,6 +1157,7 @@ module Net
       def body_fld_param
         return if NIL?
         param = {}
+        shift_token if @token.symbol == T_SPACE
         lpar
         name = case_insensitive__string; SP!; param[name] = string
         while SP?
--- a/test/net/imap/fixtures/response_parser/body_structure_responses.yml
+++ b/test/net/imap/fixtures/response_parser/body_structure_responses.yml
@@ -433,6 +433,77 @@
         4383638 NIL (\"attachment\" (\"filename\" \"test.xml\")) NIL NIL) \"mixed\"
         (\"boundary\" \"001a1137a5047848e405157ddaa3\") NIL))\r\n"
 
+  test_bodystructure_extra_space:
+    :response: "* 1 FETCH (UID 1 BODYSTRUCTURE (((\"text\" \"plain\"  (\"charset\" \"UTF-8\") NIL
+      NIL \"7bit\" 409 7 NIL NIL NIL)(\"text\" \"html\"  (\"charset\" \"UTF-8\") NIL NIL \"7bit\"
+      592 10 NIL NIL NIL) \"alternative\" (\"boundary\" \"--==_mimepart_5277b641dcc3_57d5887e8325d8\"
+      \"charset\" \"UTF-8\") NIL NIL) \"mixed\" (\"boundary\" \"--==_mimepart_5277b64110f79_57d5887e832634\"
+      \"charset\" \"UTF-8\") NIL NIL))\r\n"
+
+    :expected: !ruby/struct:Net::IMAP::UntaggedResponse
+      name: FETCH
+      data: !ruby/struct:Net::IMAP::FetchData
+        seqno: 1
+        attr:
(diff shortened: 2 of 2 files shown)


lib/net/imap/response_parser.rb
```diff
@@ -1157,6 +1157,7 @@ def body_fields
       def body_fld_param
         return if NIL?
         param = {}
+        shift_token if @token.symbol == T_SPACE
```
nevans, line 1160: Do we know if the extra space can also appear before `body-fld-param` when it is `NIL`?

For idiomatic usage, I introduced single character methods to simplify this common pattern (they also allow us to sometimes short-circuit the lexer for a significant performance improvement).  `SP!` is roughly the same as `match(T_SPACE)`, `SP?` is roughly the same as `accept(T_SPACE)`, and we also have `peek_SP?` and `lookahead_SP?`.  See [def_char_matchers](https://github.com/ruby/net-imap/blob/a32cd98673c0292494990aa7662867b13107e036/lib/net/imap/response_parser/parser_utils.rb#L17).

And, most relevant here, I started creating methods or method aliases with a "`quirky_`" prefix, simply to document that we _know_ we're deviating from the spec.  For this, I would use `quirky_SP?`: https://github.com/ruby/net-imap/blob/a32cd98673c0292494990aa7662867b13107e036/lib/net/imap/response_parser.rb#L996-L1000

So, with all of that in mind, what do you think about the following?
```suggestion
        quirky_SP? # See comments on test_bodystructure_extra_space
        return if NIL?
        param = {}
```

test/net/imap/fixtures/response_parser/body_structure_responses.yml
```diff
@@ -433,6 +433,77 @@
         4383638 NIL (\"attachment\" (\"filename\" \"test.xml\")) NIL NIL) \"mixed\"
         (\"boundary\" \"001a1137a5047848e405157ddaa3\") NIL))\r\n"
 
+  test_bodystructure_extra_space:
```
nevans, line 436: Would you mind adding a comment to document this in the test fixture?

```suggestion
  test_bodystructure_extra_space:
    :comments: |
      [GH-271] Responses from some IMAP servers contained extra space like this.
      The fix has been used since 2016, seems to be easy, and doesn't seem to
      cause any harm.
```

I just copied over some of your issue description, but please feel free to rewrite this to whatever you think would be useful.  Even better, can can share _which_ servers (the name of an ISP or email host, or a hostname, or a product name and version) need this workaround?  Thanks!

nevans (changes requested):
> Since ages we have some 'imap hacks' implemented on top of net-imap. These hacks were introduced to cope with some unusual responses from IMAP servers of our customers. I thought that I'd try to 'backport' some of them if you'd consider them viable/useful.

That's great!  My experience is very similar.  :)  And, I hope the jump from 0.3.7 to 0.4.10 isn't _too_ difficult.  As you've probably seen, some of the biggest parser changes were in the `BODYSTRUCTURE` code.  And I have even more patches that I _still_ haven't integrated into net-imap, yet!

> I'm not sure if the fix is idiomatic according to the recent changes in net-imap

That's no problem.  The updates are quite simple.  I added a couple of suggestions.  Thanks!

nevans:
I'll add that I do have some concerns about simply reading in extra spaces in general, because I have seen (or maybe I only misremembered seeing?) servers mistakenly send nothing rather than `NIL`, which converts `" NIL "` into `"  "`.  But since you've run this since 2016 with no noticeable issues (and it passes all other tests), I think it should be safe.

lib/net/imap/response_parser.rb
gaynetdinov: Thanks for clarifying the idiomatic way of doing things in net-imap! 
Unfortunately I'm not aware if we had such cases when extra space was there when body-fld-param was nil, we don't have such cases documented anywhere. 
I'm gonna commit your suggestion.

test/net/imap/fixtures/response_parser/body_structure_responses.yml
gaynetdinov: I tried to find any traces of why we've added this workaround, but couldn't find anything. I will update your description as I found that originally this workaround was introduced in 2014. So it's somewhat reasonable why I cannot find any traces of why we have this :)

nevans: That's all right.  Thanks for looking into it.

nevans (approved):

PR #276, opened 2024-05-02
💥 Drop ruby 2.7 and 3.0 support, and require 3.1

nevans:
Ruby 2.7 EOL was 2023-03-31.
Ruby 3.0 EOL was 2024-04-23.

Currently, net-imap remains compatible with ruby 2.7.  But some of my unmerged branches do use ruby 3.1 syntax (e.g: updated pattern matching and endless method definitions).  It will be nice if I don't need to update those branches for compatibility with EOL rubies.

----

_**NOTE:** The next release after this is merged should be v0.5.0._

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -7,7 +7,7 @@ jobs:
     uses: ruby/actions/.github/workflows/ruby_versions.yml@master
     with:
       engine: cruby
-      min_version: 2.7
+      min_version: 3.1
 
   build:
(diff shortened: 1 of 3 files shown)


nevans:
@shugo @hsbt Are you okay with bumping the minimum ruby version for the 0.5.x releases?

shugo (dismissed):
That sounds reasonable :+1:

hsbt (dismissed):
👌

PR #280, opened 2024-05-18
📦 Update workflow with configure_trusted_publisher

nevans:
I ran `configure_trusted_publisher` to regenerate and configure trusted publishing, and copied over most of what it created.  The only significant difference is that it adds `step-security/harden-runner`.  The other changes are minor: renaming the workflow file and the GitHub environment, using SHA1 checksums for the actions rather than tags, etc.

`configure_trusted_publisher` originally configured the trusted publisher on rubygems.org _without_ any reference to the GitHub environment.  But then I made the following PR: rubygems/configure_trusted_publisher#9, and used that to generate a new RubyGems Trusted Publisher config.


nevans:
@shugo I think this failed for `v0.4.12` because I accidentally configured the environment to require a _branch_ matching `v*`.  I've updated it to work with any _tags_ matching `v*`.  So hopefully we'll be more lucky with `v0.4.13` and `v0.5.0`. 🙂

Issue #288, opened 2024-05-24
Add Net::IMAP::Config class for global and client-scoped config


nevans:
### Motivation

The primary motivation is to simplify backward compatibility concerns while still allowing deprecation and breaking changes.  Users should be able to opt-in to new (incompatible) behaviors before they become the default, opt-out of new (incompatible) behaviors after they become the default, and load defaults by version number (similarly to how Rails handles upgrades via [Rails::Application::Configuration#load_defaults](https://edgeguides.rubyonrails.org/configuring.html#versioned-default-values)).

The secondary motivation is to scope all config options.  It should be possible to globally update defaults for client options and method parameters.  But client configuration should be able to override any global configuration.  A simple mechanism for customizing configuration inheritance should be available.

Another motivation is to avoid a proliferation of configuration methods on Net::IMAP, both global config attributes (class methods) and client config attributes (instance methods).  Currently, the number of config options is low, but that number will go up to handle options for backward and forward compatibility.

An implementation goal, not originally in this proposal but discovered during #291: Basic type checking and coercion should be simple.  For example, the timeouts are already being assigned using `@foo_timout = Integer(foo_timout)`, so the config options for them should do that too.

#### Non-goals

The following were considered, but are _not_ goals for this issue (maybe in future PRs):
* client-specific configuration (such as `host` and `port`).
* Reading config from ENV, Thread locals, Fiber locals, or Fiber storage.
* Thread safety (config is *usually* only changed up-front or during client construction)
* Ractor safety

### Proposal

`Net::IMAP::Config` will have attribute readers/writers for all config options.

All config objects (except for the default config) will inherit from a `#parent` config object.  All configs will have a hard-coded (frozen) `Config.default` as their final ancestor, and `Config.default` will be the only config with `parent.nil?`.  `Net::IMAP.config` will be a global config (not frozen), which inherits from `default`.

Every client will create its own unique Config in `#initialize`, and all unknown keyword args will be forwarded to the client config attributes.  A `config` keyword arg will set the client's _parent_ config.  This would allow a client to use defaults without inheriting from the (mutuble) global config, or to inherit from a custom ENV or Fiber.storage based config, which might itself inherit from the global config.  With no further customization, the config chain is `client -> global -> default`.

[...]

PR #291, opened 2024-06-10
Add Config class for `debug`, `open_timeout`, and `idle_response_timeout`

nevans:
This enables:
* per-client overrides for `Net::IMAP.debug`
* global defaults for `open_timeout` and `idle_response_timeout`

This feels like overkill for just these three config values, but it's written in the context of dozens of other future config options as indicated by #288.  And this should make adding new config options as simple as adding new `attr_accessor` (and default values) to the Config class.  Regarding the rest of the proposal in #288, this implements the basic config inheritance described there.  Note that the backward compatibility `load_defaults` mentioned by #288 has _not_ been implemented for this PR.

Two other significant implementation details, not originally mentioned in the #288 proposal:
* basic type checking and coercion has been added.  This happened as part of converting the timeouts to use the config: they were already being assigned using `Integer(input)`, so I preserved that.
* instead of using simple `attr_accessor`, which was my original plan, this PR delegates to an internal `Struct`.  This is partly a performance optimization (to avoid object shape proliferation), and partly because it simplifies some of the inheritance code.

### DIFF
--- a/lib/net/imap.rb
+++ b/lib/net/imap.rb
@@ -740,14 +740,12 @@ module Net
     # Returns the global Config object
     def self.config; Config.global end
 
-    # Returns the debug mode.
-    def self.debug
-      return @@debug
-    end
+    # Returns the global debug mode.
+    def self.debug; config.debug end
 
-    # Sets the debug mode.
+    # Sets the global debug mode.
     def self.debug=(val)
-      return @@debug = val
+      config.debug = val
     end
 
     # The default port for IMAP connections, port 143
--- a/lib/net/imap.rb
+++ b/lib/net/imap.rb
@@ -2609,8 +2609,6 @@ module Net
     PORT = 143         # :nodoc:
     SSL_PORT = 993   # :nodoc:
 
-    @@debug = false
-
     def start_imap_connection
       @greeting        = get_server_greeting
       @capabilities    = capabilities_from_resp_code @greeting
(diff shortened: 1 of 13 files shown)


nevans:
@shugo any objections to this?  Honestly, it feels a bit overengineered to me, for just three variables with global defaults and client overrides.  But, I'm hoping the inheritance flexibility will pay itself off as more config options are added.

I've updated #97 (with #293) and added #294, based on this config implementation.  The only remaining piece to complete #288 is a `Config#load_defaults(version)`.

Also, this is the last significant feature I want for the 0.4 branch.  With this in place, we can choose to backport deprecation warnings, with or without any new feature or behavior.  For example, I want the default config in v0.6 to disallow all implicit use of `RawData`, but that requires big updates to search (because currently complex searches need to be sent as a string).  Those big updates will _not_ be backported to 0.4, but I may backport the option to enable the deprecation warnings: it would still be useful to get warnings (and maybe convert that code to _explicitly_ use `RawData`).

shugo:
@nevans I'm not against adding `Config`, but why not just `require` it instead of `autoload`?

nevans:
> @nevans I'm not against adding `Config`, but why not just `require` it instead of `autoload`?

Good question.  Config will always be loaded. 😆  I switched it to use `require`.

nevans:
@shugo I _did_ switch it to `require_relative`... but I forgot to _push_ the version where I switched it (and a few other very minor documentation tweaks).  I'll push that later today.

nevans:
Pushed the autoload/require_relative switch in #295.

PR #300, opened 2024-06-21
✨ Add Config methods: `#to_h`, `#update`, and `#with`

nevans:
These methods were originally used to implement `#load_defaults` (to finish #288), and it seems reasonable to add them to the public API.

Unlike `SequenceSet` (which I want to have a full API to match Set, Range, and Array), I want to be careful in adding any extra public methods to `Config`.  But I think these three are basic enough to be permitted.

In other words, `#update` does the same thing that `#initialize` does with its kwargs.

I _originally_ made `#with` create a _copy_ (using `dup`), which meant it was a "sibling" to the receiver, not a child.  I did it that way because that's what I wanted for creating the versioned defaults (see #302) and I extracted the method from that code.  Also, that's how `Data#with` works and I had copied some of the rdoc from there.

But when I went back to add the tests, it seemed clear to me that the Principle of Least Surprise would expect `#with` to create a child config, inheriting from its creator.  I had already created the _instance_ method `#new` for that purpose, but `#with` is usefully different enough to justify its distinct existence (in my opinion)

### DIFF
--- a/lib/net/imap/config.rb
+++ b/lib/net/imap/config.rb
@@ -143,8 +143,26 @@ module Net
       # If a block is given, the new config object is yielded to it.
       def initialize(parent = Config.global, **attrs)
         super(parent)
-        attrs.each do send(:"#{_1}=", _2) end
+        update(**attrs)
         yield self if block_given?
+      end
+
+      # :call-seq: update(**attrs) -> self
+      #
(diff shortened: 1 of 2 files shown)


nevans:
I've also implemented `#deconstruct_keys`, `#inspect`, and `#pretty_print`.  Maybe I'll add those another PR.

PR #313, opened 2024-08-27
🐛 Fix #send_data to send DateTime as time

taku0:
`DateTime` is a subclass of `Date`.  So it was sent as date.


nevans (approved):
Aha!  I should have added a test for that.

PR #315, opened 2024-08-31
🐛 Fix #header_fld_name to handle quoted strings correctly

taku0:
## Summary

If a header field name is quoted, `ResponseParser#header_fld_name` drops the closing quotation mark.

## Reprodusing code

```ruby
parser = Net::IMAP::ResponseParser.new

parser.instance_variable_set("@str", %Q|"CONTENT-LOCATION")|)
parser.instance_variable_set("@pos", 0)
parser.instance_variable_set("@lex_state", Net::IMAP::ResponseParser::EXPR_BEG)

parser.send(:header_fld_name)
```

Expected: `"\"CONTENT-LOCATION\""`
Actual: `"\"CONTENT-LOCATION"`

## Cause

`header_fld_name` extract the raw encoded value using `@str` and `@pos`.  `header_fld_name` expects `astring` to lookahead a space or a closing parenthesis but if the field name is quoted, it calls `string` method and it doesn't lookahead a token.

## Fix

Adjust the range to extract depending on whether a token is looked ahead or not.

### DIFF
--- a/lib/net/imap/response_parser.rb
+++ b/lib/net/imap/response_parser.rb
@@ -1339,7 +1339,8 @@ module Net
         assert_no_lookahead
         start = @pos
         astring
-        @str[start...@pos - 1]
+        end_pos = @token ? @pos - 1 : @pos
+        @str[start...end_pos]
       end
 
       # mailbox-data    =  "FLAGS" SP flag-list / "LIST" SP mailbox-list /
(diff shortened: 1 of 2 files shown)


nevans (changes requested):
Good catch.  Thanks!  I'll merge and release it on Monday or Tuesday.

Could you add a simple test case to `test/net/imap/fixtures/response_parser/fetch_responses.yml`?  If you add the test case without any "expected" value, the test runner will print out the actual parsed result, for you to inspect and copy/paste into the yaml.  Thanks!

taku0:
I have amended the commit with a test case.

nevans (approved):

nevans:
@taku0 v0.4.16 has been released.  Thanks again!

PR #328, opened 2024-09-24
✅ Setup simplecov

nevans:
#327 and bugs like it can be embarrassing! 😞  I really thought I had a test for that (maybe I did, and it got lost during a rebase or refactoring?).

Adding simplecov will make it easier to hunt these sorts of bugs down and prevent them.


nevans:
WTH GitHub?  I didn't close this?

PR #329, opened 2024-09-25
✨ Add support for VANISHED responses

nevans:
`VANISHED` responses are required by two different extensions: `QRESYNC` and `UIDONLY`.

* Add a new response data type: `VanishedData`, which wraps `SequenceSet` and the `EARLIER` modifier.
* Update the parser to handle the `VANISHED` response => `VanishedData`
* Update `#expunge` and `#uid_expunge` to return `VANISHED` _(without the `EARLIER` modifier)_ responses, when they are present.
  * There is a TODO note on this: For consistency, if either of the above extensions are enabled, we should return an empty VanishedResponse when the commands have no result.  Currently, an empty array is returned.

Although neither extension is completely supported yet, they can both be (partially) used in a way that doesn't crash... so I think this is a reasonable chunk to extract into its own PR, rather than wait for a PR for either of them to be 100% ready.

### DIFF
--- a/lib/net/imap.rb
+++ b/lib/net/imap.rb
@@ -1889,22 +1889,46 @@ module Net
       send_command("UNSELECT")
     end
 
+    # call-seq:
+    #   expunge -> array of message sequence numbers
+    #   expunge -> VanishedData of UIDs
+    #
     # Sends an {EXPUNGE command [IMAP4rev1 §6.4.3]}[https://www.rfc-editor.org/rfc/rfc3501#section-6.4.3]
-    # Sends a EXPUNGE command to permanently remove from the currently
-    # selected mailbox all messages that have the \Deleted flag set.
+    # to permanently remove all messages with the +\Deleted+ flag from the
+    # currently selected mailbox.
+    #
+    # Returns either an array of expunged message <em>sequence numbers</em> or
+    # (when the appropriate capability is enabled) VanishedData of expunged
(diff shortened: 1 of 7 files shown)


nevans:
Note that this PR is based on #330.

lib/net/imap.rb
```diff
@@ -1889,18 +1889,39 @@ def unselect
...
-      synchronize do
-        send_command("EXPUNGE")
-        clear_responses("EXPUNGE")
-      end
+      expunge_internal("EXPUNGE")
     end
 
+    # call-seq:
+    #   uid_expunge -> array of message sequence numbers
+    #   uid_expunge -> VanishedData of UIDs
```
nevans, line 1923:
```suggestion
    #   uid_expunge(uid_set) -> array of message sequence numbers
    #   uid_expunge(uid_set) -> VanishedData of UIDs
```

lib/net/imap.rb
```diff
@@ -3261,6 +3282,21 @@ def enforce_logindisabled?
       end
     end
 
+    def expunge_internal(...)
+      synchronize do
+        send_command(...)
+        vanished_array = extract_responses("VANISHED") { !_1.earlier? }
+        if vanished_array.empty?
+          clear_responses("EXPUNGE")
```
nevans, line 3290: similar to `search_internal` with `SEARCH` vs `ESEARCH` responses: always clear the `EXPUNGE` responses.  They will be unusable with outdated and misleading information anyway, as soon as any `VANISHED` responses are processed.

PR #333, opened 2024-09-26
✨ Add basic ESearch support

nevans:
Parse `ESEARCH` into `ESearchResult`, with support for generic RFC4466 syntax and RFC4731 `ESEARCH` return data.  For compatibility, `ESearchResult#to_a` returns an array of integers (sequence numbers or UIDs) whenever any `ALL` result is available.

Gather ESEARCH response to #search/#uid_search.  If the server returns both `ESEARCH` and `SEARCH`, both are cleared from the responses hash, but only the `ESEARCH` is returned.

When the server doesn't send any search responses and return options are passed, return an empty ESearchResult.  This empty result will have the appropriate `tag` and `uid` values, but no `data`.  Otherwise return an empty `SearchResult`.

Add keyword param for search return options, and process it differently from `criteria`.  Also extract a `return` argument out of the `criteria` array, and process it like the `return` kwarg.  (FYI: I discovered https://bugs.ruby-lang.org/issues/20956 while working on this).

### TODO list

- [x] `ESearchResponse < Data.define(:tag, :uid, :data)`
  - [x] `#data` is a generic Array of `[name, value]` pairs.  Not using `Hash` because some names can be repeated, and the order is significant _even between different names_.  `data.assoc(name)` can be used for quick lookup.
    - [x] TODO: test cases for unknown extensions
    - [x] TODO: document that `ExtensionData` will be used for unknown return data
  - [x] Methods for compatibility with `SearchResult`:
    - [x] `#to_a` returns numbers from `#all`, `#partial`, or an empty array
    - [x] `#modseq`
  - [x] Methods for `ESEARCH` return data: `#min`, `#max`, `#all`, `#count`
  - [x] Methods for `ESEARCH` + `CONDSEQ`: `#modseq`
- [x] Parse `ESEARCH` into `ESearchResponse`
  - [x] Return data for unknown extensions: use `ExtensionData.new(tagged_ext_value)`
  - [x] Return data for `ESEARCH` and `CONDSTORE` (`MIN`, `MAX`, `ALL`, `COUNT`, `MODSEQ`): parse as Integer or SequenceSet
- [x] Support in `#search` and `#uid_search`
  - [x] Support for return options
    - [x] `return` kwarg: must be an Array, and process differently than `criteria`.
    - [x] extract `return` from array `criteria` _and process differently than `criteria`_
    - [x] When server doesn't return any response but return options are used, return an empty `ESearchResponse` rather than an empty `SearchResponse`.
    - [x] Document supported return options
  - [x] Gather `ESEARCH` responses in addition to `SEARCH` responses.  If server sends both, prefer to return the `ESEARCH` responses.
- [x] For consistency, add `charset` kwarg to `#search`, `#uid_search`

**TODO:** Remove the following from this PR (make new PRs for them):
- [x] `ESearchResponse` methods and data classes
  - [x] remove `#to_a` returning numbers from `#partial`
  - [x] remove `#partial`
    - [x] remove `ESearchResponse::PartialResult`
  - [x] remove `#updates`
  - [x] remove `#relevancy`
- [x] Parse `PARTIAL`, `ADDTO`, `REMOVEFROM`, `RELEVANCY` as generic `ExtensionData`

### Related future PRs:
- Support for `PARTIAL` (RFC9394, RFC5267)
- Support for `CONTEXT=*` (RFC5267)
  - Automatically update a specified `SequenceSet`?
  - Register a special updates response handler?
- Support for `ESORT` (RFC5267)
- Support for `SEARCH=FUZZY` (RFC6203)
  - Method to zip `#relevancy` scores with numbers from `#all` or `#partial`
  - Unsorted ESearchResult `#to_a`, `#all`, `#partial`
- SequenceSet methods:
  - unsorted `#numbers`
  - More methods for accessing or mutating SequenceSet without affecting the order
  - `#prepend` - like `#append` but adds to the beginning
  - `#insert(index, value)`
  - `#remove(index_or_range)`
  - unsorted version of `#[]`, `#at`, `#slice`
  - unsorted `#zip` with array of numbers (for `RELEVANCY`)
  - unsorted `#zip` with another SequenceSet (for `COPYUID`)

### DIFF
--- a/lib/net/imap.rb
+++ b/lib/net/imap.rb
@@ -3277,6 +3277,10 @@ module Net
         esearch = false
       in [_, Array[RETURN_WHOLE, _, *] | RETURN_START]
         raise ArgumentError, "conflicting return options"
+      in [_, Array[RETURN_WHOLE, _, *]] # workaround for https://bugs.ruby-lang.org/issues/20956
+        raise ArgumentError, "conflicting return options"
+      in [_, RETURN_START]              # workaround for https://bugs.ruby-lang.org/issues/20956
+        raise ArgumentError, "conflicting return options"
       in [return_opts, keys]
         return_opts = convert_return_opts(return_opts)
         esearch = true
--- a/lib/net/imap.rb
+++ b/lib/net/imap.rb
@@ -2020,7 +2020,9 @@ module Net
     # ==== Argument translation
     #
     # [+return+ options]
-    #   Must be an Array.  Return option names are strings.
+    #   Must be an Array.  Return option names may be either strings or symbols.
+    #   +Range+ elements which begin and end with negative integers are encoded
+    #   for use with +PARTIAL+--any other ranges are converted to SequenceSet.
     #   Unlike +criteria+, other return option arguments are not automatically
     #   converted to SequenceSet.
     #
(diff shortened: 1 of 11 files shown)


nevans:
FWIW: I discovered in testing that Yahoo does _not_ return ESEARCH results when `RETURN (PARTIAL 1:500)` is used, contrary to the PARTIAL RFC _that was written by Yahoo engineers!_

PR #334, opened 2024-10-07
✨ New config option to return frozen dup from `#responses`

nevans:
Add `:frozen_dup` option to `Config#responses_without_block`.  When set to `:frozen_dup`, `IMAP#responses` called without a type or a block will return a frozen duplicate or the responses hash.  The hash's array values will also be frozen dups.  None of the values in those arrays will be frozen, but writing to those should be uncommon.  (Net::IMAP itself will never update any response structs.)

Because `responses(type)` is relatively new and has always raised an exception, we can update it to _always_ return a frozen dup array without breaking backward compatibility.  This is done _regardless_ of the config option.  The config name (`config.responses_without_block`) is a little misleading now, but it's kept for backward compatibility.  `config.responses_without_args` was added as an alias.

### DIFF
--- a/lib/net/imap.rb
+++ b/lib/net/imap.rb
@@ -2497,17 +2497,21 @@ module Net
     private_constant :RESPONSES_DEPRECATION_MSG
 
     # :call-seq:
+    #   responses(type) -> frozen array
     #   responses       {|hash|  ...} -> block result
     #   responses(type) {|array| ...} -> block result
     #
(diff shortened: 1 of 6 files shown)


shugo:
> @shugo Any thoughts on this? I originally thought it would be better to always crash, but I can't remember why I thought that. Returning a frozen copy will still crash with some code, but it will work just fine for many others.

I agree with the proposed change.  Thank you.

PR #344, opened 2024-10-30
Make simplecov-json as optional dependency

hsbt:
In `ruby/ruby` repository, we can't use `simplecov-json` for `make test-bundled-gems`.

https://github.com/ruby/ruby/actions/runs/11569147158/job/32202320346#step:14:163

I added workaround for that. I'm happy to release v0.5.1 with this.


nevans (approved):
👍🏻

I originally considered only loading simplecov if an ENV var is present.  I like this better.  🙂

PR #345, opened 2024-10-30
✨ Enable parenthesized lists in search criteria

nevans:
This affects `#search`, `#uid_search`, `#sort`, `#uid_sort`, `#thread`, and `#uid_thread#`.

Prior to this, sending a parenthesized list in the search criteria for any of these commands required the use of strings, which are converted to `RawData`, which has security implications with untrusted inputs.

With this change, arrays will only be converted into `SequenceSet` when _every_ element in the array is a valid `SequenceSet` input.  Otherwise, the array will be left alone, which allows us to send parenthesized lists without using strings and `RawData`.

For example, some searches this change enables:
* combining criteria to pass into `OR`, `NOT`, `FUZZY`, etc.
  * `search(["not", %w(flagged unread)])` converts to:
    `SEARCH not (flagged unread)`
* Adding return options (we should also add a `return` kwarg). 
  * `uid_search(["RETURN", ["PARTIAL", 1..50], "UID", 12345..67890]` converts to:
    `UID SEARCH RETURN (PARTIAL 1:50) UID 12345:67890`
  * _Note that `PARTIAL` supports negative ranges, which can't be coerced to SequenceSet.
    They'll need to be sent as strings, for now._
  * _Note that searches with return options should return `ESEARCH` results, which are currently unsupported.
    See #333._

This should be backward compatible: previously these inputs would simply raise an exception.

### DIFF
--- a/lib/net/imap.rb
+++ b/lib/net/imap.rb
@@ -1949,9 +1949,13 @@ module Net
     #     * +Range+
     #     * <tt>-1</tt> and +:*+ -- both translate to <tt>*</tt>
     #     * responds to +#to_sequence_set+
-    #     * nested +Array+
+    #     * +Array+, when each element is one of the above types, a positive
+    #       +Integer+, a sequence-set formatted +String+, or a deeply nested
+    #       +Array+ of these same types.
     #   * Any +String+ is sent verbatim when it is a valid \IMAP atom,
     #     and encoded as an \IMAP quoted or literal string otherwise.
+    #   * Any other nested +Array+ is encoded as a parenthesized list, to group
+    #     multiple search keys (e.g., for use with +OR+ and +NOT+).
     #   * Any other +Integer+ (besides <tt>-1</tt>) will be sent as +#to_s+.
     #   * +Date+ objects will be encoded as an \IMAP date (see ::encode_date).
     #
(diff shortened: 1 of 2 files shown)


lib/net/imap.rb
```diff
@@ -1989,17 +2013,23 @@ def uid_expunge(uid_set)
     #
     # ===== Capabilities
     #
-    # If [CONDSTORE[https://www.rfc-editor.org/rfc/rfc7162.html]] is supported
+    # When +IMAP4rev2+ is enabled, or when the server supports +IMAP4rev2+ but
+    # not +IMAP4rev1+, ESearchResult is always returned instead of SearchResult.
+    #
```
nevans, line 2018: This PR was rebased/cherry-picked from #333, which adds ESearchResult for `ESEARCH` responses.  But this PR doesn't have support for that yet.  The docs should be updated to reflect the current behavior.

```suggestion
```

lib/net/imap/sequence_set.rb
```diff
@@ -276,16 +276,29 @@ class SequenceSet
...
+      # (Enumerable other than Array or Set) will be rejected.
+      Coercible = ->(obj) do
+        case obj
+        when SequenceSet            then true
+        when Integer, Range, *STARS then true
+        when String                 then REGEXP.match?(obj.b)
+        when Array, Set             then obj.all?(Coercible) && !obj.empty?
+        else                             obj.respond_to?(:to_sequence_set)
+        end
+      end
```
nevans, line 301: 1. This is really only going to be used for search criteria, not for any other command argument types, so the name should reflect that.
2. `Set` isn't used for anything other input type in Net::IMAP, and I don't expect it ever will be.  So we should always interpret `Set` as coercible.
3. On the other hand, `Range` _may_ be used in ways that are incompatible with SequenceSet.  See, e.g: `RETURN (PARTIAL -50:-1)`.  That complicates things a little bit: the recursive pattern will need to behave differently from the top-level pattern. 🙁
4. We don't need to check if the Array is empty.  An empty parenthesized list and an empty sequence set are both invalid, so it's simpler and better to let `SequenceSet.[]` (or `SequenceSet#validate`) raise an exception than to send an invalid `()` to the server.

```suggestion
      CoercibleSearchArg = ->(obj) do
        case obj
        when SequenceSet, Set       then true
        when Integer, *STARS        then true
        when String                 then REGEXP.match?(obj.b)
        when Array                  then obj.all?(CoercibleSearchArgMember)
        else                             obj.respond_to?(:to_sequence_set)
        end
      end
      
      CoercibleSearchArgMember = ->(member) { obj in Range | Coercible }
      private_constant :CoercibleMember
```

nevans: I somehow forgot/missed that `Range` and `-1` _are_ converted to SequenceSet in the original.  So, at least for backward compatibility, we need to still convert those.

nevans: For handling negative ranges, like in the `PARTIAL` extension, we could try one of:
1. Only convert `Range` into `SequenceSet` when it's bounded by positive integers, -1, or nil (end-less/begin-less).
2. We only see negative ranges in the return options, so if we add a new `return` kwarg, we don't need to worry about making `["PARTIAL", -50..-1]` work in the `criteria` arg.

nevans: And `sequence-set` formatted strings can't be sent as atoms, since they might contain `*` (an atom special), so they wouldn't be properly encoded as `astring` search key args.

E.g: `search(%w[SUBJECT *])` should produce `SEARCH SUBJECT "*"`, not `SEARCH SUBJECT *`.

PR #367, opened 2024-12-20
✨ Add support for `PARTIAL` extension (RFC9394)

nevans:
- Updates `#search`, `#uid_search` with the `PARTIAL` return option which adds `ESearchResult#partial` return data.
 - Updates `#uid_fetch` with the `partial` modifier (kwarg).

### DIFF
--- a/lib/net/imap.rb
+++ b/lib/net/imap.rb
@@ -2424,12 +2424,12 @@ module Net
     # {[RFC7162]}[https://tools.ietf.org/html/rfc7162] in order to use the
     # +changedsince+ argument.  Using +changedsince+ implicitly enables the
     # +CONDSTORE+ extension.
-    def fetch(set, attr, mod = nil, changedsince: nil)
-      fetch_internal("FETCH", set, attr, mod, changedsince: changedsince)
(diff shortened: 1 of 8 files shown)


lib/net/imap.rb
```diff
@@ -2403,19 +2432,42 @@ def uid_search(...)
...
-    #   uid_fetch(set, attr, changedsince: nil) -> array of FetchData
+    #   uid_fetch(set, attr, changedsince: nil, partial: nil) -> array of FetchData
     #
     # Sends a {UID FETCH command [IMAP4rev1 §6.4.8]}[https://www.rfc-editor.org/rfc/rfc3501#section-6.4.8]
     # to retrieve data associated with a message in the mailbox.
     #
     # Similar to #fetch, but the +set+ parameter contains unique identifiers
     # instead of message sequence numbers.
     #
+    # When #uid_fetch may also be given a +partial+ range, which can be used to
```
nevans, line 2448:
```suggestion
    # On capable servers, #uid_fetch may also be given a +partial+ range, which can be used to
```

Issue #287, opened 2024-05-24
Flaky test with macOS platform


hsbt:
I sometimes got the following failure with `ruby/ruby` and `make test-bundled-gems`.

```
Testing the net-imap gem
/Users/hsbt/Documents/github.com/ruby/ruby/exe/ruby -C /Users/hsbt/Documents/github.com/ruby/ruby/gems/src/net-imap /Users/hsbt/Documents/github.com/ruby/ruby/.bundle/bin/rake test
Loaded suite /Users/hsbt/Documents/github.com/ruby/ruby/.bundle/gems/rake-13.2.1/lib/rake/rake_test_loader
Started
/Users/hsbt/Documents/github.com/ruby/ruby/.bundle/gems/test-unit-3.6.2/lib/test/unit/fixture.rb:284: warning: the block passed to 'priority_setup' defined at /Users/hsbt/Documents/github.com/ruby/ruby/.bundle/gems/test-unit-3.6.2/lib/test/unit/priority.rb:183 may be ignored
F
====================================================================================================================================================================================
Failure: test_imaps_with_ca_file(IMAPTest):
  Exception raised:
  <#<OpenSSL::SSL::SSLError: SSL_read: tlsv1 alert unknown ca (SSL alert number 48)>>
  Backtrace:
    /Users/hsbt/Documents/github.com/ruby/ruby/.ext/common/openssl/buffering.rb:80:in 'OpenSSL::SSL::SSLSocket#sysread'
    /Users/hsbt/Documents/github.com/ruby/ruby/.ext/common/openssl/buffering.rb:80:in 'OpenSSL::Buffering#fill_rbuff'
    /Users/hsbt/Documents/github.com/ruby/ruby/.ext/common/openssl/buffering.rb:236:in 'OpenSSL::Buffering#gets'
    /Users/hsbt/Documents/github.com/ruby/ruby/gems/src/net-imap/lib/net/imap.rb:2720:in 'Net::IMAP#get_response'
    /Users/hsbt/Documents/github.com/ruby/ruby/gems/src/net-imap/lib/net/imap.rb:2629:in 'Net::IMAP#receive_responses'
    /Users/hsbt/Documents/github.com/ruby/ruby/gems/src/net-imap/lib/net/imap.rb:2606:in 'block in Net::IMAP#start_receiver_thread'.
[...]

PR #389, opened 2025-01-30
Omit flaky test with macOS platform

hsbt:
For https://github.com/ruby/net-imap/issues/287

`OpenSSL::SSL::SSLError` is raised at 13/69 times in my macOS box. I always retried all of test suite of `ruby/ruby` everyday.

I know we should resolve this with root causes. But I have not enough time to investigate this.

And I got the following timeout issue with 4 times. 

```
#<Thread:0x00000003236f3850 /Users/hsbt/Documents/github.com/ruby/ruby/gems/src/net-imap/test/net/imap/fake_server/test_helper.rb:10 run> terminated with exception (report_on_exception is true):
/Users/hsbt/Documents/github.com/ruby/ruby/lib/timeout.rb:40:in 'Timeout::Error.handle_timeout': execution expired (Timeout::Error)
```

I extend that timeout value to handle that.

I didn't get test failed running with ten times after applying above two workarounds.

### DIFF
--- a/test/net/imap/fake_server/test_helper.rb
+++ b/test/net/imap/fake_server/test_helper.rb
@@ -4,7 +4,7 @@ require_relative "../fake_server"
 
 module Net::IMAP::FakeServer::TestHelper
 
-  def run_fake_server_in_thread(ignore_io_error: false, timeout: 5, **opts)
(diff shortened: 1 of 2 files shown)


nevans (approved):

nevans:
@hsbt This is reasonable.

PR #400, opened 2025-02-06
✨ Add AppendUIDData and CopyUIDData classes

nevans:
It isn't possible to convert `UIDPlusData` to use `SequenceSet` internally, while still maintaining full backwards compatibility.  So two new classes were added (`AppendUIDData` and `CopyUIDData`).  The intention is to remove `UIDPlusData` in `v0.6.0`.

* Both `AppendUIDData` and `CopyUIDData`:
  * uid-set attributes are `SequenceSet` objects (not arrays of integers)
  * new validations have been added to the initializers
  * `data.size` to get the uid-set size
* Only `CopyUIDData`:
  * `uid_mapping => source_uid_to_assigned_uid_hash`
    (This behaves identically to the existing `UIDPlusData#uid_mapping`.)
  * `copyuid.assigned_uid_for(source_uid) => assigned_uid`
    `copyuid[source_uid] => assigned_uid`
  * `copyuid.source_uid_for(assigned_uid) => source_uid`
  * `copyuid.each_uid_pair` _(aliases: `each_pair`, `each`)_

Although they are not perfectly compatible, upgrading from `UIDPlusData` to `CopyUIDData` and `AppendUIDData` should be relatively straightforward.

This PR only adds the classes, but they are unused.  For backward compatibility, a parser config option is needed in order to replace `UIDPlusData` with them.  That was split into a separate PR (mostly just to simplify the release notes generation):
* #401

### DIFF
--- a/lib/net/imap/response_data.rb
+++ b/lib/net/imap/response_data.rb
@@ -8,6 +8,7 @@ module Net
     autoload :SearchResult,     "#{__dir__}/search_result"
     autoload :SequenceSet,      "#{__dir__}/sequence_set"
     autoload :UIDPlusData,      "#{__dir__}/uidplus_data"
+    autoload :AppendUIDData,    "#{__dir__}/uidplus_data"
     autoload :VanishedData,     "#{__dir__}/vanished_data"
 
     # Net::IMAP::ContinuationRequest represents command continuation requests.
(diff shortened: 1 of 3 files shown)


nevans:
@hoffi Since you submitted the original code, I assume this may affect you.  Any thoughts?  (See also: #401)

hoffi:
Looks fine to me 👍

PR #401, opened 2025-02-06
🔧 Add parser config for `APPENDUID`/`COPYUID`, 🗑️ Deprecate UIDPlusData

nevans:
* Depends on #400

Adds `Net::IMAP::Config#parser_use_deprecated_uidplus_data` attribute.  This config attribute causes the parser to use the new AppendUIDData and CopyUIDData classes instead of CopyUIDData.

`AppendUIDData` and `CopyUIDData` are _mostly_ backward-compatible with `UIDPlusData`.  Most applications should be able to upgrade with no changes.

The plan is to remove `UIDPlusData` in `v0.6`.  To simplify upgrades, the config option can stay until `0.7`.  But, after `UIDPlusData` has been removed, maybe only allow setting it to `false` .

### DIFF
--- a/lib/net/imap/config.rb
+++ b/lib/net/imap/config.rb
@@ -287,6 +287,32 @@ module Net
       #
       # Alias for responses_without_block
 
+      # Whether ResponseParser should use the deprecated UIDPlusData or
+      # CopyUIDData for +COPYUID+ response codes, and UIDPlusData or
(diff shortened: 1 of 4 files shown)


nevans:
@hoffi Since you submitted the original code, I assume this may affect you.  Any thoughts?  (See also: #400)

hoffi:
Looks fine to me 👍

PR #433, opened 2025-04-18
♻️ Extract ResponseReader from `get_response`

nevans:
It's nice to extract a little bit of the complexity from the core `Net::IMAP` class.  But my primary motivation was so that I could directly test this code quickly and in isolation from needing to simulate a full IMAP connection.

Note that, because it wraps `@sock`, the ResponseReader _must_ be recreated (or updated) when `@sock` is replaced for TLS (or, in the future, for the `COMPRESS` extension).

### DIFF
--- a/lib/net/imap/response_reader.rb
+++ b/lib/net/imap/response_reader.rb
@@ -1,0 +1,38 @@
+# frozen_string_literal: true
+
+module Net
+  class IMAP
+    # See https://www.rfc-editor.org/rfc/rfc9051#section-2.2.2
+    class ResponseReader # :nodoc:
+      attr_reader :client
+
+      def initialize(client, sock)
+        @client, @sock = client, sock
+      end
+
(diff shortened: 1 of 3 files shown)


lib/net/imap/response_reader.rb
```diff
@@ -0,0 +1,38 @@
...
+
+module Net
+  class IMAP
+    # See https://www.rfc-editor.org/rfc/rfc9051#section-2.2.2
+    class ResponseReader # :nodoc:
+      attr_reader :client
+
+      def initialize(client, sock)
+        @client, @sock = client, sock
+      end
```
nevans, line 11: This version of the `ResponseReader` isn't actually using `client` yet, so I could've left that out of this PR.  But upcoming PRs will need the client for passing to callbacks (and also for its config).

nevans:
:twi

PR #463, opened 2025-04-29
⚡️ Improve `SequenceSet#xor` performance by ~2x

nevans:
Obviously, the performance improvement is highly dependant on what data you're using, whether YJIT is enabled, etc.  I saw results ranging from 1.5x faster to 2.7x faster.  The benchmark script is included.

For a benchmark run using sets with 10k members:
```
new impl   79.061 (± 8.9%) i/s   (12.65 ms/i) -    392.000 in   5.004322s
old impl   32.736 (±15.3%) i/s   (30.55 ms/i) -    162.000 in   5.052839s
```
The old implementation was ~2.42x slower.

For a benchmark run using very sparse sets with 100 members:
```
new impl   4.295k (±13.5%) i/s  (232.81 μs/i) -     21.476k in   5.102536s
old impl   2.459k (±11.3%) i/s  (406.69 μs/i) -     12.095k in   5.000148s
```
This time, the old implementation was ~1.75x slower.

I have some other (much bigger) PRs that should give even bigger performance (and memory use) improvements, but this is simple and effective.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -15,6 +15,7 @@ gem "test-unit"
 gem "test-unit-ruby-core", git: "https://github.com/ruby/test-unit-ruby-core"
 
 gem "benchmark-driver", require: false
+gem "benchmark-ips",    require: false
 
 group :test do
   gem "simplecov",        require: false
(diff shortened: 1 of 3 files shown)


nevans:
I didn't have proper tests over `#xor` and had accidentally pasted in an alternate implementation for `#union`!  D'oh!

PR #470, opened 2025-05-05
✅ Add TruffleRuby to CI (and JRuby?)

### LINKED ISSUE #454: Add JRuby and TruffleRuby to CI
I'd like to add CI for JRuby and TruffleRuby, but the last time I tried I got a bunch of errors (admittedly, I didn't spend much time trying to figure them out). 

Ideally, we could fix the errors without significant changes, and `net-imap` could be fully functional under those implementations.

If the changes needed for full compatibility are significant, maybe we can just mark some tests as pending.  That will at least serve as a warning of what's incompatible, and prevent new changes from inadvertently widening that gap.

nevans:
Fixes #454. (at least, it will fix the TruffleRuby part... once it's passing)

Cc: @eregon


eregon:
I had a quick look through the failures at https://github.com/ruby/net-imap/actions/runs/14711726747/job/41285563407?pr=470

```
Failure: test_recursive_inspect(Net::IMAP::TestData)
/home/runner/work/net-imap/net-imap/test/net/imap/test_data_lite.rb:180:in `test_recursive_inspect'
     177:         # anonymous class
     178:         list = klass[value: 1, tail: [2, 3, 4]]
     179:         seen = "#<data #{klass.inspect}:...>"
  => 180:         assert_equal(
     181:           "#<data value=1, head=nil," \
     182:           " tail=#<data value=2, head=#{seen}," \
     183:           " tail=#<data value=3, head=#{seen}," \
<internal:core> core/throw_catch.rb:36:in `catch'
<internal:core> core/throw_catch.rb:36:in `catch'
<"#<data value=1, head=nil, tail=#<data value=2, head=#<data #<Class:0x411e8>:...>, tail=#<data value=3, head=#<data #<Class:0x411e8>:...>, tail=#<data value=4, head=#<data #<Class:0x411e8>:...>, tail=nil>>>>"> expected but was
<"#<data value=1, head=, tail=#<data value=2, head=#<data #<Class:0x411e8>:...>, tail=#<data value=3, head=#<data #<Class:0x411e8>:...>, tail=#<data value=4, head=#<data #<Class:0x411e8>:...>, tail=>>>>">
```
That looks like an issue of Data#inspect or #to_s in truffleruby not calling inspect on the values.
But https://github.com/oracle/truffleruby/blob/ac88a0fe68bf957f75af7d316594b89731fdec4e/src/main/ruby/truffleruby/core/data.rb#L141-L152 seems to clearly call `inspect` on each value, so it's a mystery (unless something redefines `nil.inspect`? I would hope not, but would be worth checking).

For the other failures I don't see anything obvious, I think they'll need investigation with some debugging, I suggest to do that later and so omit these 9 failures for now.

lib/net/imap/data_lite.rb
```diff
@@ -27,7 +27,12 @@
 
 module Net
   class IMAP
-    data_or_object = RUBY_VERSION >= "3.2.0" ? ::Data : Object
+    data_or_object =
+      if RUBY_VERSION >= "3.2.0" && !%w[truffleruby jruby].include?(RUBY_ENGINE)
```
eregon, line 31: How about `defined?(::Data) && ::Data.respond_to?(:define)` ?
This should be more accurate, notably TruffleRuby 24.0 and more recent do have `Data.define`.

I suspect it might even fix some CI failures to correctly use Data on TruffleRuby.

eregon: Ah the problem is a stack overflow with the builtin Data: https://github.com/ruby/net-imap/actions/runs/14684658586/job/41211746960
Skipping that `test_recursive_inspect` test is probably better though, to keep using the optimized core Data (which should be fixed to handle recursion in `inspect`).

test/net/imap/test_regexps.rb
```diff
@@ -4,6 +4,7 @@
 require "test/unit"
 
 return unless Regexp.respond_to?(:linear_time?)
+return if %w[truffleruby jruby].include?(RUBY_ENGINE)
```
eregon, line 7: `Regexp.linear_time?` is implemented on TruffleRuby, and from https://github.com/ruby/net-imap/actions/runs/14687111151/job/41217223575 it didn't seem to cause any failure.
JRuby seems to crash before reaching that anyway.
So I think this commit can be removed.

headius: JRuby will not crash before reaching this once we look into failures. We do implement `linear_time?` but it always returns false right now because we have not had resources to implement the linear time optimizations in joni. Not sure how that impacts this test.

lib/net/imap/data_lite.rb
nevans: @eregon, thanks.  Yeah... I often get tripped when I try to do too much with `define?`!  I'll push a version with `defined?(::Data) && ::Data.respond_to?(:define)`.

test/net/imap/test_regexps.rb
nevans: Thanks.  I replaced the guards at the top of the file with this: 5862712779
```ruby
  def test_linear_time(data)
    assert Regexp.linear_time?(data.regexp)
  rescue NoMethodError
    raise if RUBY_ENGINE == "ruby"
    pend "Regexp.linear_time? not implemented in #{RUBY_ENGINE} #{RUBY_ENGINE_VERSION}"
  rescue Test::Unit::AssertionFailedError
    raise if RUBY_ENGINE == "ruby"
    pend "Regexp might backtrack in #{RUBY_ENGINE} #{RUBY_ENGINE_VERSION}"
  end
```

So now, all them are marked as pending for JRuby (I tested both 9.4.12.0 and 10.0.0.0), and 23 out of 99 are marked as pending for TruffleRuby 24.2.0.

eregon: Nice. It'd be super convenient to also print the Regexp in the `pend` message to easily see what it looks like.

nevans: @eregon the regexp const name is printed in the failure (because it's used as the name for each row of `data`).  But I agree: for quicker inspection, seeing the actual regexp in the error would be helpful.

eregon:
> Both JRuby and TruffleRuby still fail in various ways when inheriting from Data or Struct.

One thought I had given this gem 3.1+ would be to use Struct (which exists since forever) instead of Data (+ freeze it, BTW thanks for [this issue](https://github.com/oracle/truffleruby/issues/3850), should be an easy quick fix).
Then you wouldn't need DataLite at all and wouldn't need to test that Data|DataLite work the same/good enough for the usage here.

But you're saying there are other issues with Struct too?

nevans:
> One thought I had given this gem is 3.1+ would be to use Struct (which exists since forever) instead of Data

That's fair.  I _did_ consider going that route.  But I also wanted the more constrained API of Data vs Struct.

I originally started down the path of building a facade over Struct.  And then a switched from that to just a PORO.  But the API I wanted turned out to be the Data-API.  So in the end, I just used Data.  And I'd already done much of the work for several features using Data... but I was waiting for v0.6.0 (which will require ruby > 3.2) before merging them.

`FetchData` is a good example of where the `Struct` API comes into direct conflict with the natural (Principle of Least Surprise) API for a `FETCH` response object: `#[]` should delegate to `attr#[]`, and `#size` should delegate to `attr["RFC822.SIZE"]`.  My long-term goal for many (not all) of the existing structs is to add deprecation warnings to `Struct`-specific methods for a couple of versions (i.e: a couple of years), then swap out their implementations for `Data` or POROs.

So I was originally going to hold off on merging _any_ new data structures until v0.6.0, just to avoid these issues.  But work priorities bumped up the schedule for #333 and #329, which were both features I'd already mostly completed on my own time, based on `Data`.  So I had to decide between 1) using a fork for ~18 months, 2) converting the objects to `Struct`, 3) converting the objects to a Struct-facade (probably just a subset of the `Data` API, or 4) using a `Data` polyfill which would only be used for one (now-EOL) version of ruby.  🤷🏻  (In the end we used a fork in production for almost three months before I merged and released everything.)

> (+ freeze it, BTW thanks for [this issue](https://github.com/oracle/truffleruby/issues/3850), should be an easy quick fix).

great! 👍🏻 

> Then you wouldn't need DataLite at all and wouldn't need to test that Data|DataLite work the same/good enough for the usage here.

My intention has always been (see https://github.com/ruby/net-imap/pull/352#issuecomment-2466295656) for v0.6.0 to delete `DataLite` and set `Net::IMAP::Data = ::Data`.  It only exists for 1) backwards compatibility with a single (now-EOL) version of ruby, and 2) YAML encoding.  And the next release of `psych` will add YAML encoding: ruby/psych#692. 🎉

While working on this ticket, I briefly considered keeping it in v0.6.0, just for TruffleRuby and JRuby... but... I'm not gonna do that.  😉  I expect the JRuby and TruffleRuby bugfixes to come faster than my 0.6.0 release plans.  🙂

> But you're saying there are other issues with Struct too?

Yep.  If you don't get to it first, I'll open TruffleRuby issue for it later today.  But the failing `net-imap` tests are here: 7fd8e2c03.

And, to me, they seem very similar to the Data issues.  My intuition may be completely wrong, but I would not be surprised if the fixes for `Data` also fix `Struct`, or vice versa.

nevans:
> And, to me, they seem very similar to the Data issues. My intuition may be completely wrong, but I would not be surprised if the fixes for `Data` also fix `Struct`, or vice versa.

On further investigation, I suspect the Struct issue is not inheritance related, but a difference in how Struct handles mixed positional and keyword arguments.

I created oracle/truffleruby#3855.

nevans:
@eregon here's the issue I saw with `IO#gets(CRLF, limit)`:
```ruby
#!/usr/bin/env ruby

puts ?=*72
puts "Ruby Engine:   #{RUBY_ENGINE} #{RUBY_ENGINE_VERSION}"

require 'bundler/inline'
gemfile do
  gem "stringio", ">= 3.1.7"
end
require "stringio"
puts "Using StringIO #{StringIO::VERSION}"
io = StringIO.new("123456789\r\n123456789\r\n")
pp stringio: io.gets("\r\n", 10)

r, w = IO.pipe
w.write "123456789\r\n123456789\r\n"
w.close
pp io_pipe: r.gets("\r\n", 10)

__END__
$ ruby test.rb && RBENV_VERSION=truffleruby-24.2.1 ruby test.rb
========================================================================
Ruby Engine:   ruby 3.4.2
Using StringIO 3.1.7
{stringio: "123456789\r"}
{io_pipe: "123456789\r"}
========================================================================
Ruby Engine:   truffleruby 24.2.1
Using StringIO 3.0.1
{:stringio=>"123456789\r\n"}
{:io_pipe=>"123456789\r"}
```

I didn't make a TruffleRuby ticket for this.  I'm guessing this is already a known issue, given that it silently failed to use a newer version of `StringIO` (even after I manually gem installed it).  And it looks like it only affects the test which uses StringIO and probably doesn't affect real IO.

duerst:
> My long-term goal for many (not all) of the existing structs is to add deprecation warnings to `Struct`-specific methods for a couple of versions (i.e: a couple of years), then swap out their implementations for `Data` or POROs.

Sorry for the side question, but can you explain what a PORO is? I wasn't able to find this with a Web search, probably just my bad.

eregon:
PORO = Plain Old Ruby Object (might derive from POJO use in Java)

nevans:
All of the currently failing JRuby tests were passing for me locally.  Oh well.  Maybe it's some sort of race condition when calling `Socket#close` to shutdown the FakeServer.

But, `IO#close` shouldn't raise an exception due to "stream closed in another thread", should it?  I thought that exception was only raised on threads that are blocked on the closed FD.  If I'm reading the CRuby code correctly (that's a _very_ big "if"), `rb_io_close_m` 1) first checks to see if it's already been closed, and 2) doesn't drop the GVL until after it's been closed (e.g: `fptr_finalize_flush` is called with `KEEPGVL`).

eregon:
All the TruffleRuby issues are fixed now, except https://github.com/oracle/truffleruby/issues/3858 where for AGE it's not currently planned to be fixed.
The CI on this PR is also already passing on TruffleRuby.
Could we maybe merge this (maybe after some cleanup as there is no need to detect working Data anymore) and not test on JRuby yet in CI since it's failing?

nevans:
> All the TruffleRuby issues are fixed now, except [oracle/truffleruby#3858](https://github.com/oracle/truffleruby/issues/3858) where for AGE it's not currently planned to be fixed. The CI on this PR is also already passing on TruffleRuby.

Excellent!  Great work.  🙂 

> Could we maybe merge this (maybe after some cleanup as there is no need to detect working Data anymore) and not test on JRuby yet in CI since it's failing?

Yes!  I'll clean this up and merge it (or a subset of it) before the 0.5.9 release (sometime this week).

PR #502, opened 2025-08-07
⚡🔎 Abridged SequenceSet#inspect output

nevans:
When using SequenceSet to store mailbox state for mailboxes with many millions of messages, the last thing you want to do is accidentally generate a string and output that!

So this establishes a maximum number of entries to output as an inspect string, currently 512.  Above that number, we output only a much smaller number of entries from the head and tail, currently 16 from each end.  When the sequence set has an internal denormalized string, we use regexps to grab the first and last groups of entries from that.

```ruby
Net::IMAP::SequenceSet(((1..5000) % 2).to_a).inspect
#=> #<Net::IMAP::SequenceSet 2500 entries "1,3,5,7,9,11,13,15,17,19,21,23,25,27,29,31,...(2468 entries omitted)...,4969,4971,4973,4975,4977,4979,4981,4983,4985,4987,4989,4991,4993,4995,4997,4999">
```

### A note on the implementation

At first, I used a simple Regexp match anchored to `\z` to grab the tail entries from the string.  But that was shockingly slow even when the string isn't extremely large.

Next I benchmarked replacing `Regexp#match` with `String#rpartition` and that performed fine, using the same naive regexps.  `String#rindex` was also fine.  They are both slower than `Regexp#match` anchored to `\A`, but not too bad and not dramatically slower based on string size.  I found myself wishing for a `String#rscan` or `String#rsplit`.  😆

Then I benchmarked replacing `\d+` with `\d{0,10}` or `[1-9]\d{1,9}`, and those both made the regexp fast... but `Regexp.linear_time?` doesn't like nested quantifiers.  Fully expanding both the `\d+` and the `(entry){100}` keeps it fast _and_ satisfies `Regexp.linear_time?`.

But that regexp is UGLY! 🙁

### DIFF
--- a/lib/net/imap/sequence_set.rb
+++ b/lib/net/imap/sequence_set.rb
@@ -396,6 +396,23 @@ module Net
       STARS     = [:*, ?*, -1].freeze
       private_constant :STARS
 
+      INSPECT_MAX_LEN      = 512
+      INSPECT_TRUNCATE_LEN =  16
+      private_constant :INSPECT_MAX_LEN, :INSPECT_TRUNCATE_LEN
+
+      # /(,\d+){100}\z/ is shockingly slow on huge strings.
+      # /(,\d{0,10}){100}\z/ is ok, but ironically, Regexp.linear_time? is false.
+      #
+      # This unrolls all nested quantifiers.  It's much harder to read, but it's
+      # also the fastest out of all the versions I tested.
+      nz_uint32   = /[1-9](?:\d(?:\d(?:\d(?:\d(?:\d(?:\d(?:\d(?:\d(?:\d)?)?)?)?)?)?)?)?)?/
+      num_or_star = /#{nz_uint32}|\*/
+      entry       = /#{num_or_star}(?::#{num_or_star})?/
+      entries     = ([entry] * INSPECT_TRUNCATE_LEN).join(",")
+      INSPECT_ABRIDGED_HEAD_RE = /\A#{entries},/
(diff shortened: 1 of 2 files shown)


lib/net/imap/sequence_set.rb
```diff
@@ -396,6 +396,23 @@ class SequenceSet
...
+      INSPECT_MAX_LEN      = 512
+      INSPECT_TRUNCATE_LEN =  16
+      private_constant :INSPECT_MAX_LEN, :INSPECT_TRUNCATE_LEN
+
+      # /(,\d+){100}\z/ is shockingly slow on huge strings.
+      # /(,\d{0,10}){100}\z/ is ok, but ironically, Regexp.linear_time? is false.
+      #
+      # This unrolls all nested quantifiers.  It's much harder to read, but it's
+      # also the fastest out of all the versions I tested.
+      nz_uint32   = /[1-9](?:\d(?:\d(?:\d(?:\d(?:\d(?:\d(?:\d(?:\d(?:\d)?)?)?)?)?)?)?)?)?/
```
nevans, line 408: n.b:
```ruby
(2**32 - 1).digits.count =>
  10
/[1-9]#{9.times.inject(//) { /(?:\d#{_1.source})?/ }.source}/ ==
  /[1-9](?:\d(?:\d(?:\d(?:\d(?:\d(?:\d(?:\d(?:\d(?:\d)?)?)?)?)?)?)?)?)?/ =>
  true
```

PR #528, opened 2025-10-03
✅ Add TruffleRuby to CI

nevans:
No application code is updated by this PR; only the test suite.  3 tests are marked pending, but the rest are all passing.  So, with a few caveats, it looks like TruffleRuby _mostly_ works.  This doesn't mean that TruffleRuby is "officially supported", but it _does_ mean that we'll try not to merge PRs that break tests that aren't broken.

Partially replaces #454.


nevans:
@eregon FINALLY!  😉

eregon:
:tada:

PR #543, opened 2025-10-16
🔥 Use psych (>= 5.2.5) for encoding Data objects

nevans:
Depends on #541, which drops the `Data` polyfill.

Unlike #541, this is a breaking change, because objects that would previously encode with a `ruby/object:Net::IMAP:DataSubclass` tag will instead encode with a `ruby/data:Net::IMAP:DataSubclass` tag.  Any YAML with the old tag will create an (effectively) empty data object: the data members will not be initialized (but ivars will be set).


nevans:
@eregon It looks like TruffleRuby doesn't like the `Data` deserialization that was added to psych 5.2.5 (ruby/psych#692).  For `!ruby/data` tagged objects, `Psych::Visitors::ToRuby` calls `#init_struct` which calls `rb_struct_initialize`, and that symbol understandably isn't found for TruffleRuby.  I used that in ruby/psych#692 because that's how CRuby's Marshal handles both Struct and Data objects.

I guess TruffleRuby will either need to support a compatible `rb_struct_initialize`, or it will need to patch `Psych::Visitors::ToRuby#init_struct`?

eregon:
Could you file an issue at https://github.com/truffleruby/truffleruby/issues ?
Obviously we should add TruffleRuby in ruby/psych CI to catch earlier in the future.

PR #547, opened 2025-11-09
🔍 Improve Config#pretty_print (for `Kernel::pp`)

nevans:
Viewing config objects in Irb was _much_ to messy to easily inspect: it printed every attribute for each ancestor, and most of them would simply be the Net::IMAP::Config::AttrInheritance::INHERITED const, repeated over and over.

So, just like `Config#inspect` was improved by #546, this simplifies the `pp` output to make it more useful for inspection on the console.

Unlike `Config#inspect`, `Config#pretty_print` prints _every_ config attribute, including attributes that have been defined by a named default config.

### DIFF
--- a/lib/net/imap/config.rb
+++ b/lib/net/imap/config.rb
@@ -523,11 +523,41 @@ module Net
       #     #      inherits from Net::IMAP::Config.global debug=true
       #     #      inherits from Net::IMAP::Config.default>"
       #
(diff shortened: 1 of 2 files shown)


nevans:
Tests are only failing in windows for ruby-head due to:
* ruby/ruby#14866
* ruby/test-unit-ruby-core#19
* ruby/test-unit-ruby-core#20

Those fixes should all be merged shortly.

Issue #484, opened 2025-06-11
Improve `SequenceSet` memory use and performance


nevans:
`SequenceSet` is very useful for managing mailbox state.  But I've known since first implementing it that, even though it uses binary searches for `O(lg n)` in many places, the current naive implementation of `SequenceSet` has some serious performance issues, especially at larger scales (e.g: when a set holds millions of members, which is very common for "All Mail" type mailboxes).  For many scenarios, it performs worse than simply keeping a sorted array of integers (using `Array#bsearch`, etc).

Most notably: by storing each element as a `[min, max]` range/run tuple, we create a 8-byte pointer to 40-byte array for each `uint32x2` run.  We could easily use 1/3rd as much memory, either by storing a flat array (runs represented by `flat.each_cons(2)`) or two "columnar" arrays (runs represented by `mins.zip(lens)`.  And that could be cut in half by storing the `uint32` values in a packed string or by merging each `uint32x2` run into a single `uint64` Integer—in practice this will almost always be less than `UINT62_MAX` and thus stored as a `FIXNUM`.

So we're basically using 6x more memory than needed, and that memory isn't even in a contiguous block.  And that's before we get into more advanced tricks, such as splitting a set into `uint16` chunks, and then dynamically encoding each chunk—as an array of runs or numbers, or even as a bitset.  (While working on this, I accidentally re-invented a naive version of [Roaring Bitmaps](https://roaringbitmap.org/)... and when I started searching for prior art, I discovered [roaring bitmaps](https://roaringbitmap.org)!)

And that's all before we get into more advanced tricks, for example: splitting large sets into `UInt16` chunks, allowing runs to be stored as `uint16x2`.  And once that's done, we can then dynamically encode each chunk as either run length encoded, a bitset, or simple array of numbers.  I've tested the encoding of real-life mailbox UID sets, and this last step provides _enormous_ memory savings.

So, I always wanted to change the naive array-of-arrays implementation.

Additionally, I'd also like to add less naive versions of specific set operations.  Most set operations are implemented as simply `dup.mutating_version!(other)`.  In my benchmarks, some fairly straightforward tricks can be employed to `AND` and `XOR` operations 20% to 300% faster for real-world scenarios (i.e: my own production code).

And, of course, none of this should be done without good benchmarks to justify the significant jump in complexity.

PR #574, opened 2025-12-10
🔖 Release 0.6

nevans:
My plan is to release v0.6.0 on Sunday.

There are two remaining blockers:
* [x] #540
  This requires only a few tweaks to the current PR, so the _config options_ can have their own deprecation period.  They'll raise a warning in the parser when they would've triggered.  This is related to deprecation and deletion, so it really is a blocker for v0.6.0.
* [ ] #484
  This won't be _done_, but I did want to at least move off of the array of arrays approach, which uses 6x more memory than necessary!  My current run length encoded branch is significantly faster on _most_ benchmarks... except for operations that index into it as if it were an array of numbers: `#at`, `#slice`/`#[]`, `#find_index`, etc.  These are all already `O(n)` in the _current_ implementation... but dealing with the encoded runs slows them down even more (at least in the current implementation).  And, for how I use `SequenceSet`, I really need these operations operations to be significantly _faster_ than they already are.

  If I'm not happy with my (simple) "chunked RLE" sequence set implementation by Sunday, we'll just ship what we already have.

I did also hope to finish the QRESYNC code, and I'll probably even be working on that this week and next for my employer.  But I don't think I'll have time to make the `net-imap` parts of that work release ready.  Everything else can wait for v0.6.1 or v0.7.0.

### DIFF
--- a/lib/net/imap.rb
+++ b/lib/net/imap.rb
@@ -788,7 +788,7 @@ module Net
   # * {IMAP URLAUTH Authorization Mechanism Registry}[https://www.iana.org/assignments/urlauth-authorization-mechanism-registry/urlauth-authorization-mechanism-registry.xhtml]
   #
   class IMAP < Protocol
-    VERSION = "0.6.0-dev"
+    VERSION = "0.6.0"
 
     # Aliases for supported capabilities, to be used with the #enable command.
     ENABLE_ALIASES = {


nevans:
I didn't finish the migration from `SequenceSet` implemented as an array of arrays into one of the other (faster, 3x-6x less memory use) because my preferred approach is still too slow for indexing/ranking by number (which is critically important for mapping UIDs to sequence numbers!).

But although it's a major refactoring, it should be 100% compatible... so it can be in the 0.6.1 release... and maybe still be released in time for ruby 4.0??? 🤞🏻

nevans:
I can't fix this right now, but I'll get it pushed in a couple hours.

PR #593, opened 2026-01-02
Delete test/net/imap/test_data_lite.rb

nobu:
The target of this test file has been delete by #543.

_Originally posted by @nobu in https://github.com/ruby/net-imap/issues/543#issuecomment-3704422702_
> This deleted `lib/net/imap/data_lite.rb`, but `test/net/imap/test_data_lite.rb` that was added alongside it still exists.


nevans:
Thanks, @nobu.  Sorry for the delay.

PR #629, opened 2026-03-25
Handle deep response recursion as ResponseParseError

Masamuneee:
This changes `ResponseParser#response` to rescue `SystemStackError` and re-raise `Net::IMAP::ResponseParseError` with the message "response recursion too deep".

### DIFF
--- a/lib/net/imap/response_parser.rb
+++ b/lib/net/imap/response_parser.rb
@@ -691,6 +691,8 @@ module Net
         CRLF!
         EOF!
         resp
+      rescue SystemStackError
+        parse_error("response recursion too deep")
       end
 
       # RFC3501 & RFC9051:
--- a/test/net/imap/test_response_parser.rb
+++ b/test/net/imap/test_response_parser.rb
@@ -356,4 +356,53 @@ class ResponseParserTest < Net::IMAP::TestCase
     assert_instance_of Net::IMAP::CopyUIDData, response.data.code.data
   end
 
+  def deeply_nested_bodystructure(depth)
+    leaf = '("TEXT" "PLAIN" NIL NIL NIL "7BIT" 1 1)'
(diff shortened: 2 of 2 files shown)


Masamuneee:
cc @nevans

nevans (dismissed):
This looks good to me, as-is.  But, to ensure the tests run reliably on all platforms, we may want to mix in the deterministic aspects of #630 (see my comments there).

nevans:
@Masamuneee to get this passing quickly for truffleruby, you can either wrap the tests with `omit_if_truffleruby`, or use some of the strategy from #630.

Masamuneee:
I followed the strategy from #630 instead of using `omit_if_truffleruby` because I think it gives a better long-term test.

The main reason is that `omit_if_truffleruby` would only hide the CI failure on one runtime, while still leaving the test dependent on actual stack depth everywhere else. That makes the test less stable and less portable, and it also drops coverage on TruffleRuby entirely.

With the current approach, the production change still happens in `response`, as suggested, and the test remains deterministic by patching the recursion entry point (`body`, `thread_data`, or `tagged_ext_val`) rather than relying on the real VM stack limit. At the same time, the input is still a real deeply nested response shape, so the test documents the kind of response that can plausibly trigger this path.

I also kept the `ENV["TEST_RESPONSE_PARSER_STACK_LEVEL"] == "original"` escape hatch so the same test can still be exercised against the real recursive behavior when desired, without making CI depend on stack depth.

So the goal here was:
  - keep the implementation aligned with your suggested `response` rescue
  - keep the test deterministic across runtimes
  - still use realistic response examples
  - avoid skipping coverage on TruffleRuby unless absolutely necessary

nevans:
@Masamuneee Thanks.  That allows it to pass on TruffleRuby now.

nevans (approved):

PR #630, opened 2026-03-25
Wrap SystemStackError as ResponseParseError in ResponseParser

thesmartshadow:
This makes ResponseParser#parse handle SystemStackError the same way it already handles other parse failures, by wrapping it as Net::IMAP::ResponseParseError.

The goal here is not to change the existing mitigation strategy. Parse failures should still terminate response processing and lead to disconnect behavior as they do today. This change just makes the failure mode consistent for callers, so SystemStackError does not bypass the normal ResponseParseError path.

I also added a deterministic regression test that forces SystemStackError without depending on Ruby or OS stack depth. The test asserts that parse raises ResponseParseError and preserves the original SystemStackError as its cause.

### DIFF
--- a/lib/net/imap/response_parser.rb
+++ b/lib/net/imap/response_parser.rb
@@ -691,6 +691,8 @@ module Net
         CRLF!
         EOF!
         resp
+      rescue SystemStackError => error
+        raise exception(error.message), cause: error
       end
 
       # RFC3501 & RFC9051:
--- a/test/net/imap/test_response_parser.rb
+++ b/test/net/imap/test_response_parser.rb
@@ -201,6 +201,26 @@ class ResponseParserTest < Net::IMAP::TestCase
     end
   end
 
+  test "deeply nested BODYSTRUCTURE raises ResponseParseError instead of SystemStackError" do
+    parser = Net::IMAP::ResponseParser.new
+
+    def parser.body
+      raise SystemStackError, "stack level too deep"
+    end unless ENV["TEST_RESPONSE_PARSER_STACK_LEVEL"] == "original"
+
+    depth = 10_000
+    leaf = '("TEXT" "PLAIN" NIL NIL NIL "7BIT" 1 1)'
+    bodystructure = depth.times.reduce(leaf) { |part, _| "(#{part} \"MIXED\")" }
+    response = "* 1 FETCH (BODYSTRUCTURE #{bodystructure})\r\n"
+
+    err = assert_raise(Net::IMAP::ResponseParseError) do
+      parser.parse(response)
+    end
+
+    assert_match(/stack level too deep/i, err.message)
+    assert_instance_of(SystemStackError, err.cause)
+  end
+
   def assert_deprecated_appenduid_data_warning
     assert_warn(/#{__FILE__}.*warning.*parser_use_deprecated_uidplus_data is ignored/) do
       result = yield


lib/net/imap/response_parser.rb
```diff
@@ -38,6 +38,12 @@ def parse(str)
         @lex_state = EXPR_BEG
         @token = nil
         return response
+      rescue SystemStackError => error
+        wrapped = exception(error.message)
+        if config.debug?
+          warn wrapped.detailed_message(parser_state: true, parser_backtrace: true)
+        end
+        raise wrapped, cause: error
```
nevans: I think I'd prefer to wrap the exception in `response` and re-raise.  That way, the debug handling doesn't need to be duplicated.

test/net/imap/test_response_parser.rb
```diff
@@ -201,6 +201,21 @@ def test_fetch_binary_and_binary_size
     end
   end
 
+  test "SystemStackError is wrapped as ResponseParseError" do
+    parser = Net::IMAP::ResponseParser.new
+
+    def parser.response
+      raise SystemStackError, "stack level too deep"
+    end
```
nevans: Normally I want to avoid patching the object under test, but in this case I'm okay with doing it to ensure this always passes, regardless of what the stack size is or any other future improvements.

_But_...
1. If you rescue and re-raise in `#response`, then `#response` can't be the method we patch.
2. I'd strongly prefer to use real example responses to _document_ what kind of response might conceivably raise this exception.  By using real live examples, we demonstrate the bug better.  Ideally, we could just comment out the patched method (or set an environment variable) to test it "for real".

So, can we update this to:
1. Use real examples that really do raise SystemStackError, at least on your computer.
2. Patch the method that initiates the recursive loop, e.g: `#body`, `#thread_data`, `#tagged_ext_val`.

e.g:
```ruby
  test "deeply nested BODYSTRUCTURE raises ResponseParseError instead of SystemStackError" do
    parser = Net::IMAP::ResponseParser.new

    def parser.body
      raise SystemStackError, "stack level too deep"
    end unless ENV["TEST_RESPONSE_PARSER_STACK_LEVEL"] == "original"
     
    depth = 10_000
    leaf = '("TEXT" "PLAIN" NIL NIL NIL "7BIT" 1 1)'
    bodystructure = depth.times.reduce(leaf) {|part, _| "(#{part} \"MIXED\")" }
    response = "* 1 FETCH (BODYSTRUCTURE #{bodystructure})\r\n"

    err = assert_raise(Net::IMAP::ResponseParseError) do
      parser.parse(response)
    end

    assert_match(/stack level too deep/i, err.message)
    assert_instance_of(SystemStackError, err.cause)
  end
```

nevans (changes requested):
Thanks for this!

thesmartshadow:
Updated  the wrapping now happens in `response`, so the existing debug handling stays in `parse`. I also changed the test to use a real deeply nested BODYSTRUCTURE-shaped input while patching `body` for deterministic behavior.

I re-ran the parser test and the full test suite locally, and both pass.

lib/net/imap/response_parser.rb
```diff
@@ -691,6 +691,8 @@ def response
         CRLF!
         EOF!
         resp
+      rescue SystemStackError => error
+        raise exception(error.message), cause: error
```
nevans, line 695: Note that `cause` is automatically set to `$!` (which is set to the current exception being handled).  So, generally we don't assign it explicitly unless we want a _different_ exception (e.g: raising from a different thread).

So I think this should be fine:
```suggestion
      rescue SystemStackError => error
        raise exception(error.message)
```

Alternatively, this should also capture the `#cause`, and I think I'd prefer to use our own custom error message:
```suggestion
      rescue SystemStackError
        parse_error("response recursion too deep")
```

nevans:
Your changes were good, but #629 was also updated to incorporate your changes here, and I merged that.  Were you working with @Masamuneee as a colleague (or an alt)?  Since @Masamuneee copied your testing technique, I can credit you both in the release notes.

thesmartshadow:
Thx @nevans  I appreciate that

No, @Masamuneee is not me and we were not coordinating. I independently worked on this and opened my PR based on the discussion here.

If you’d like to credit us both in the release notes, that sounds great to me  thank you fr

nevans:
@thesmartshadow That's funny.  Because @Masamuneee and I had been discussing it prior to opening the PR, and your fix was basically what I'd requested from @Masamuneee.  Am I correct in suspecting that is something that common code analysis tools are surfacing now?

Anyway, thanks for the PR!

Masamuneee:
I was also shocked that @thesmartshadow opened a PR right after mine, especially since that’s exactly what @nevans and I were discussing earlier. I just read through the PR and ended up using it myself! lol

thesmartshadow:
Thx @nevans  fair question I wasn’t coordinating with @Masamuneee; I used CodeQL to help inspect the code path, then reproduced it locally. Honestly, it is kind of funny in a good way the bug really points to one pretty natural fix, so it makes sense we landed on something very similar.

thesmartshadow:
LOL, @Masamuneee  that’s actually pretty great. Sometimes a bug just pulls everyone toward the same clean fix  glad your PR helped move it over the line

PR #642, opened 2026-04-16
⚡ Much faster ResponseReader performance

nevans:
> [!NOTE]
> This PR fixes quadratic time complexity when reading large responses containing many string literals.
> This should only be an issue when connecting to an _untrusted hostile server_ (or without TLS).

This _significant_ improves the performance of `ResponseReader`, which runs in the receiver thread.

For responses which contain many literals, **_v0.6.3's runs in O(n²) time_**!

`v0.6.3` is **1.15x to 1.5x** slower reading responses with no literals.
`v0.6.3` is **4x** slower reading 100 byte responses with four literals.
`v0.6.3` is **800x** slower reading 100KiB "pathological" responses (filled with zero-byte literals).

### 💥 Minor breaking change

_NOTE: It is not expected that this will affect any current usage.  If you **are** affected by this, please describe your scenario in a comment or a new issue._

Rather than read `config.max_response_size` from the client once (or more) per line/literal, it's now loaded from the config only once per response.  This gives a gives a greater speedup for each literal in the response.

Previously, changing `max_response_size` took effect after each line or literal is read.  They now take effect after each whole response is read.  This only makes a difference for responses that contain literals.

The config change latency was _already_ non-deterministic, prior to this change.  The simplest workaround is to simply call `#noop` after making the change, to flush the current response and reset the response reader with the new config.

### 📉 Benchmarks

The benchmarks below (behind the disclosure summaries) were run against the each of the following commits:
* 8c2fb18 — v0.6.3
* 341ab62 — "0: rindex re" - fixes dominant O(n²) issue, but slower for "no literals".
* 49c516d — "1: end_with?" - fixes "no literal" case, and always faster than v0.6.3.
* 6f82e28 — "2: get literal" - fixes secondary issue (maybe O(n²)?) with large responses.
* 0e89dbb — "3: memo config" - linear (per literal) improvement, minor breaking change
* 7e1af98 — "4: chk too big" - linear (per literal) improvement

Please note that, because the iteration counts for the following benchmarks were determined by the performance of v0.6.3, and the current version runs _so much_ faster (especially for larger responses), the current version's benchmarks benchmarks finish before they've had time to get up to speed, so those numbers are lower than they would be if the benchmark were given a longer warmup.  This is especially true for the YJIT numbers.

##### Without YJIT
<details><summary><tt>benchmark-driver benchmarks/response_reader.yml --filter "no literals" --run-duration 15</tt></summary>

```console
Warming up --------------------------------------
[...]

PR #643, opened 2026-04-17
Pass an Exception to #raise

eregon:
* Fixes `TypeError: exception class/object expected` on truffleruby because #raise with a non-Exception non-String is not supported currently on TruffleRuby.

Unfortunately it seems https://github.com/ruby/net-imap/pull/642 broke on TruffleRuby and because of https://github.com/ruby/test-unit-ruby-core/pull/23 it wasn't noticed.

In this PR I do an easy fix for [this failure](https://github.com/eregon/net-imap/actions/runs/24559483023/job/71804267846#step:5:38).
However after that the test suite still fails with a timeout.

I also tried increasing the timeout (from 10 to 100 [commit](https://github.com/ruby/net-imap/compare/master...eregon:net-imap:apply_timeout_scale)) but it did not help.
When that test is run in isolation (`bundle exec ruby -Itest/lib -rhelper test/net/imap/test_imap_max_response_size.rb`) it works fine though, so it sounds like there is some global state/side effect that makes the test fail somehow when run as part of the full test suite.

I don't have time to investigate more now, maybe I can take another look in the evening.

### DIFF
--- a/lib/net/imap/response_reader.rb
+++ b/lib/net/imap/response_reader.rb
@@ -70,11 +70,10 @@ module Net
         empty? ? 3 : done? ? 0 : (literal_size || 0) + 2
       end
 
-      def guard_response_too_large! = (raise exception if response_too_large?)
-
-      def exception(msg = nil)
-        ResponseTooLargeError.new(
-          msg, max_response_size:, bytes_read:, literal_size:,
+      def guard_response_too_large!
+        return unless response_too_large?
+        raise ResponseTooLargeError.new(
+          max_response_size:, bytes_read:, literal_size:,
         )
       end
 
--- a/lib/net/imap/response_reader.rb
+++ b/lib/net/imap/response_reader.rb
@@ -70,7 +70,7 @@ module Net
         empty? ? 3 : done? ? 0 : (literal_size || 0) + 2
       end
 
-      def guard_response_too_large! = (raise self if response_too_large?)
+      def guard_response_too_large! = (raise exception if response_too_large?)
 
       def exception(msg = nil)
         ResponseTooLargeError.new(
--- a/test/net/imap/fake_server/test_helper.rb
+++ b/test/net/imap/fake_server/test_helper.rb
@@ -16,6 +16,7 @@ module Net::IMAP::FakeServer::TestHelper
   def run_fake_server_in_thread(ignore_io_error: false,
                                 report_on_exception: true,
                                 timeout: 10, **opts)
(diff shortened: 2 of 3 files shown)


lib/net/imap/response_reader.rb
```diff
@@ -70,7 +70,7 @@ def min_response_remaining
         empty? ? 3 : done? ? 0 : (literal_size || 0) + 2
       end
 
-      def guard_response_too_large! = (raise self if response_too_large?)
+      def guard_response_too_large! = (raise exception if response_too_large?)
 
       def exception(msg = nil)
```
eregon: I don't know the intention but it sounds unusual to implement `#exception` on non-Exception objects.
It means e.g. users could do `raise someNet_IMAP_ResponseReader_instance` but I would think very few would expect to work, they would expect an error because it's raising something which isn't an Exception.

So maybe this method should be inlined in its only caller, like before #642?

nevans: Yeah, I was getting _way_ too cute with that one! 🤣  I just did it that way in passing, without thinking very much of it, and I'd completely intended to replace it before merging!  I do like the idea of relying on `#exception` duck-typing, where it makes sense _semantically_.  But... it does _not_ make semantic sense here.

nevans: Another refactoring branch (with a more substantial change to mitigate against pathological performance issues) was raising the same exception from multiple places.  I extracted that and named it `#exception` as a temporary placeholder.  But I discovered a more significant performance improvement at the same time.  So it accidentally migrated into into this branch after a few rebase/cherry-pick rearrangements.  😳

nevans (dismissed):

nevans (dismissed):
I actually ran the tests before pushing, _this_ time. 😳

nevans:
I wasn't triggering that timeout locally, but since I now know about `EnvUtil.timeout_scale` (😉), I've used it on `FakeServer`.

nevans:
Hmm.. no.  I think this may actually be due to 0e89dbb58!

nevans:
Yes, it was due to 0e89dbb.  I'd forgotten that I'd added a test for the original behavior.

And now I'm second-guessing whether I want that behavior change to go out in an `x.y.z` point release...  But I think I'm going to stick with my original rationale: the original behavior was already non-deterministic.  And, where exactness is required, the correct approach was always to either make the update it in a response handler, or simply trigger a response (i.e: with `noop`)

nevans (approved):

eregon:
Great, thank you for the fixes!

PR #691, opened 2026-06-06
🧵🥅 Reraise receiver thread errors with caller's backtrace

nevans:
Previously, when the receiver thread sent an exception to the client thread, it would be raised with the receiver thread's backtrace, which wasn't very useful.

Now a copy of the exception will be raised, and it will use the current thread's `backtrace`.  If the original has a backtrace or a cause, the original exception will be set as the copy's `cause`.

---

There's other work that should be done to normalize receiver thread exception propagation to client threads.  But that can wait for another PR (in another release).

### DIFF
--- a/test/net/imap/test_imap.rb
+++ b/test/net/imap/test_imap.rb
@@ -13,7 +13,7 @@ class IMAPTest < Net::IMAP::TestCase
 
   if defined?(OpenSSL::SSL::SSLError)
     def test_imaps_unknown_ca
-      assert_raise(OpenSSL::SSL::SSLError) do
+      assert_local_raise(OpenSSL::SSL::SSLError) do
         imaps_test do |port|
           begin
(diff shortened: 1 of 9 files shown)


test/net/imap/test_imap.rb
```diff
@@ -286,14 +288,53 @@ def test_unexpected_eof
...
   end
 
+  test "exception from response reader" do
+    with_fake_server ignore_io_error: true, ignore_abrupt_eof: true do |server, imap|
+      imap.instance_exec do
+        def @reader.read_response_buffer
+          raise StandardError, "crashy crash"
+        end
+      end
+      assert_reraised(StandardError, imap:) do
```
nevans: This looks like it might be a flaky test due to a threading race condition.  It can either raise this, or it can raise `IOError, "stream closed in another thread"`

PR #693, opened 2026-06-07
🧵Refactor receive responses thread sync

nevans:
Related to #692, but that handles code that executes in client thread(s) and this handles code that executes in the receiver thread.

Notable changes:
* The receiver loop only calls `synchronize` once per loop, to handle the response.
  * Various exception handlers and cleanup code is moved into adjacent synchronize blocks.
  * Resetting `@exception` at the start is moved to the end (`unless connection_closed`).
* The logout state is now entered _before_ signaling the conditional variables for the last time.  This is less surprising than doing it afterward, and it also makes some of the other `state_logout!` calls redundant.  `connection_state` is relatively new, and it is not expected that this difference should impact much code anyway.
* An `EOFError` may be assigned a receiver thread backtrace, similarly to other exceptions that are raised by response reading or parsing.  This will be significantly improved by #691.
* Simplified exception handling by simply letting the response reader and response parser exceptions bubble up, which allows us to assign `@exception` in the synchronize block inside the ensure clause.  This makes the loop much simpler to understand, and now we don't release and immediately re-acquire the lock with (slightly) inconsistent state.
* ~`@sock.close` for exceptions from `get_response` is no longer synchronized.  This converts some instances of `IOError, "stream closed"` into `IOError, "stream closed by another thread"`.~
  _This change was reverted._  It made the tests flakier, and it's not too burdensome to keep it in a `synchronized` block.

### DIFF
--- a/lib/net/imap.rb
+++ b/lib/net/imap.rb
@@ -3517,9 +3517,6 @@ module Net
     def receive_responses
       connection_closed = false
       until connection_closed
-        synchronize do
-          @exception = nil
-        end
         begin
           resp = get_response
         rescue Exception => e
@@ -3570,6 +3567,8 @@ module Net
             @exception = e
             @tagged_response_arrival.broadcast
             @continuation_request_arrival.broadcast
+          ensure
+            @exception = nil unless connection_closed
           end
         end
       end
--- a/lib/net/imap.rb
+++ b/lib/net/imap.rb
@@ -3518,18 +3518,12 @@ module Net
       connection_closed = false
       until connection_closed
         begin
-          resp = get_response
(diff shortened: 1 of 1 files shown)


nevans:
It looks like some of this made tests that assert specific connection closing errors a bit flakier.  On the one hand, it's not a big deal if client and receiver thread code, running simultaneously, sometimes results in `IOError, "closed in another thread"` rather than some specific error.  On the other hand... maybe we should try to avoid that inconsistency, if only for keeping the test suite simple.

nevans:
The last bit may make it into a future PR.  But the bulk of the refactoring was done via #694.

PR #695, opened 2026-06-08
♻️ Extract `handle_response` from `receive_responses`

nevans:
Extracting `handle_response` makes the `receive_responses` _much_ easier to read, IMO.  Also, this block of code has been surrounded by a redundant `begin`/`end`, since #694 pulled the `rescue` clause inside the `synchronize` block.

`connection_closed` was only assigned by this portion of the loop, so it's been converted into the inverse of the return value for `handle_response`, and the loop is now a simple `Kernel#loop`.  The `receive_responses` method returns when `handle_response` returns falsey.

### DIFF
--- a/lib/net/imap.rb
+++ b/lib/net/imap.rb
@@ -3516,47 +3516,9 @@ module Net
 
     def receive_responses
       exception = nil
-      connection_closed = false
-      until connection_closed
+      loop do
         resp = get_response or raise EOFError, "end of file reached"
-        begin
-          synchronize do
-            case resp
-            when TaggedResponse
-              @tagged_responses[resp.tag] = resp
-              @tagged_response_arrival.broadcast
-              case resp.tag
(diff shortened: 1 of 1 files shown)


lib/net/imap.rb
```diff
@@ -3516,47 +3516,9 @@ def tcp_socket(host, port)
...
-            end
-          rescue Exception => e
-            @exception = e
-            @tagged_response_arrival.broadcast
-            @continuation_request_arrival.broadcast
-          ensure
-            @exception = nil unless connection_closed
-          end
-        end
+        handle_response(resp) or return
```
nevans, line 3521: It would make sense to convert this boolean loop breaker into a `throw` or `raise`.  But the caveat is that we still need to reset `@exception` after handling an exception that _doesn't_ closing the connection.  So we'd still need to track that inside `handle_response`.

PR #706, opened 2026-06-11
Remove dead code resulting from cherry-pick of 6f82e28.

jackorp:
Cherry-pick of https://github.com/ruby/net-imap/commit/6f82e28f714ecced2396144bce0155e2cdb0256e seems to have added instead of replaced the `done?` method in the commit bfdae21.


nevans:
Thanks!  That's annoying.  Warnings are enabled during test runs, so it would have printed a warning.  But I missed it.

nevans (approved):

PR #709, opened 2026-06-29
Fix flaky test_starttls_stripping_ok_sent_before_response

hsbt:
`test_starttls_stripping_ok_sent_before_response` is failing intermittently on the `Windows 2025-vs2026 (test-bundled-gems)` runner in ruby/ruby CI. The test expects `Net::IMAP::InvalidTaggedResponseError` but instead gets `Errno::ECONNABORTED (SSL_connect)`.

These are real, reproducing failures against net-imap 0.6.4.1 bundled in ruby/ruby:

https://github.com/ruby/ruby/actions/runs/28358709394/job/84007854240
https://github.com/ruby/ruby/actions/runs/28339193097/job/83950550687

In the first run the same test failed on both the initial attempt and the retry on the same VM, so it is not a one-off hiccup. Windows 2022 and other runners pass, which points to a timing-dependent flake specific to the slower VS2026 runner.

The test verifies the defense against a STARTTLS response-injection attack. The server sends `RUBY0001 OK ...` before STARTTLS is negotiated, and `finish_sending_command` must notice the already-stored tagged response and raise `InvalidTaggedResponseError` before entering `start_tls_session`. For that check to fire, the receive thread must have parsed and stored the injected response in `@tagged_responses` first.

The old code only waited with `sleep 0.010`, which confirms the server flushed the bytes but not that the client's receive thread finished recording the response. On the VS2026 runner the storage does not complete within 10ms, so the check is bypassed, the command proceeds to `SSL_connect`, and the non-TLS test server closes the socket, raising ECONNABORTED.

This replaces the fixed sleep with an `add_response_handler` plus `Queue` synchronization, the same pattern already used by the adjacent "premature tagged OK response" test. `handle_response` stores the tagged response before invoking response handlers, so waiting for the handler to fire guarantees the value is in `@tagged_responses` before `starttls` runs. The wait is wrapped in `Timeout` so a regression that prevents the response from arriving fails fast instead of hanging.

Verified locally by running the test 50+ times without a failure; the full `test_imap.rb` suite (34 tests) also passes.

### DIFF
--- a/test/net/imap/fake_server/socket.rb
+++ b/test/net/imap/fake_server/socket.rb
@@ -51,7 +51,7 @@ class Net::IMAP::FakeServer
     def ignore_closed?(fallback)
       yield
     rescue IOError => err
-      close if !closed? && (@tcp_socket.closed? || @tls_socket.closed?)
+      close if !closed? && (@tcp_socket.closed? || @tls_socket&.closed?)
       return fallback if err.message.match?(/stream closed|closed stream/i)
       raise
     end
--- a/test/net/imap/test_imap.rb
+++ b/test/net/imap/test_imap.rb
@@ -174,6 +174,7 @@ class IMAPTest < Net::IMAP::TestCase
     def test_starttls_stripping_ok_sent_before_response
       # to coordinate between threads (better than sleep)
       server_to_client, client_to_server = Queue.new, Queue.new
+      rcvr_to_client = Queue.new
       imap = nil
       server = create_tcp_server
       port = server.addr[1]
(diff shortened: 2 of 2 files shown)


nevans:
@hsbt I hope you don't mind that I rewrote your git trailer to use the [linux kernel's `Assisted-by` format](https://github.com/torvalds/linux/blob/dc59e4fea9d83f03bad6bddf3fa2e52491777482/Documentation/process/coding-assistants.rst#L46).  I'm considering adding that request to `CONTRIBUTING.md`.

nevans (approved):

hsbt:
>I rewrote your git trailer to use the [linux kernel's Assisted-by format](https://github.com/torvalds/linux/blob/dc59e4fea9d83f03bad6bddf3fa2e52491777482/Documentation/process/coding-assistants.rst#L46)

Oh, thanks. I'm not sure why claude added that trailer instead of `Co-Authored-By:` at this time. I'll be more careful next time.

nevans:
> > I rewrote your git trailer to use the [linux kernel's Assisted-by format](https://github.com/torvalds/linux/blob/dc59e4fea9d83f03bad6bddf3fa2e52491777482/Documentation/process/coding-assistants.rst#L46)
> 
> Oh, thanks. I'm not sure why claude added that trailer instead of `Co-Authored-By:` at this time. I'll be more careful next time.

Sorry for the misunderstanding.  Claude _did_ add its usual `Co-Authored-By:` trailer to your commit.  But I strongly prefer the `Assisted-by:` format, so _I changed it_ to that.

This follows [the format required by the linux kernel's `coding-assistants.md`](https://github.com/torvalds/linux/blob/dc59e4fea9d83f03bad6bddf3fa2e52491777482/Documentation/process/coding-assistants.rst#L46):
> When AI tools contribute to kernel development, proper attribution helps track the evolving role of AI in the development process.  Contributions should include an Assisted-by tag in the following format:
> 
>     Assisted-by: AGENT_NAME:MODEL_VERSION [TOOL1] [TOOL2]
> 
> Where:
> 
> * ``AGENT_NAME`` is the name of the AI tool or framework
> * ``MODEL_VERSION`` is the specific model version used
> * ``[TOOL1] [TOOL2]`` are optional specialized analysis tools used
>   (e.g., coccinelle, sparse, smatch, clang-tidy)
> 
> Basic development tools (git, gcc, make, editors) should not be listed.
> 
> Example:
> 
>     Assisted-by: Claude:claude-3-opus coccinelle sparse

hsbt:
I understood. I will replace that trailer at next time. Thanks to your explanation.

PR #712, opened 2026-07-05
🥅 Test for NULL bytes before sending string args

nevans:
The latest release _does_ validate that atoms, quoted strings, literals don't contain any NULL bytes.  However generic string arguments would not be fully validated until they attempted to send (inside `#send_string_data`).

This change allows us to raise the exception during argument validation, rather than waiting until we are sending the command (which is too late to save the connection from being broken).

Additionally:
* Although many other tests provide coverage for generic string arguments, I discovered that we were missing any basic test coverage for sending UTF-8 string arguments.  So I added some tests to make sure we have full coverage over each case in `send_string_data`.
* I added some (internal only) rdoc to clarify the difference between `astring` (used for generic string args) and `Atom`.  This (or something like it) may be promoted to public rdoc in a future version.

### DIFF
--- a/lib/net/imap/command_data.rb
+++ b/lib/net/imap/command_data.rb
@@ -15,6 +15,9 @@ module Net
       case data
       when nil
       when String
+        if data.include?("\0")
+          raise DataFormatError, "String argument contains NULL byte"
+        end
       when Integer
         # Covers modseq-valzer, which is the largest valid IMAP integer
         if data.negative?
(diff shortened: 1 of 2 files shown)


lib/net/imap/command_data.rb
```diff
@@ -55,6 +55,10 @@ def send_data(data, tag = nil)
       end
     end
 
+    UNQUOTABLE_CHARS = /\r\n/n
```
nevans, line 58: Oops.

PR #729, opened 2026-07-31
✅🚧 Mark local stacktrace tests pending for JRuby

nevans:
The stacktraces just don't line up.

* JRuby Issue: jruby/jruby#9528
* Fixed by: jruby/jruby#9528

### DIFF
--- a/test/lib/helper.rb
+++ b/test/lib/helper.rb
@@ -210,7 +210,9 @@ class Net::IMAP::TestCase < Test::Unit::TestCase
       assert_raise(expected, &block)
     end
     stack = caller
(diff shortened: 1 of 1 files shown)


nevans:
There are still a couple of issues related to this:
* JRuby's `caller(1)` differs from `(raise rescue $!.backtrace[1..])`.  _(easy workaround)_
* There is a deeper issue of exceptions from the receiver thread are _sometimes_ not correctly conveyed to the main thread, which raises IOError instead.  This makes the tests flaky when they contain assertions about receiver thread error handling.  But I think it's less a flaky test issue and more an actual threading issue in the client code, and it'll be fixed separately.

PR #733, opened 2026-08-20
Use benchmark_driver

nobu:
benchmark-driver is an alias gem to install benchmark_driver.


nevans:
Although I don't really understand _why_ CI is failing, #736 fixes it, and this PR has nothing to do with that.

nevans (approved):

PR #738, opened 2026-08-27
Compare SearchResult values instead of hashes in eql?

OskarEichler:
## Summary

Compare ordered array contents and modseq in SearchResult#eql? instead of treating matching hash values as proof of equality. Also reject a non-nil modseq when the receiver has no modseq. The existing order-insensitive == and hash calculation are unchanged.

## Reproduction

```ruby
require 'net/imap'
a = Net::IMAP::SearchResult[1, modseq: nil]
b = Net::IMAP::SearchResult[1, modseq: 2]
p [a.eql?(b), b.eql?(a)] # before: [true, false]; after: [false, false]
colliding = Class.new(Net::IMAP::SearchResult) { def hash = 7 }
p({colliding[1, modseq: 2] => :a, colliding[9, modseq: 3] => :b}.size)
# before: 1; after: 2
```

## Verification

- 400 focused checks; 73 failing expectations before, zero afterward. They cover nil/non-nil modseq, ordered contents, plain arrays, subclasses, deliberately colliding hashes and Hash lookup. Related merged #514 fixed ==, not eql?.
- Existing `rake test` on this isolated branch: **1726 tests, 12588 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications**, Ruby 4.0.6 via rbenv. The baseline also passes all 1,726 tests; assertion counts vary slightly across runs.

## Compatibility and limits

No API, dependency or Ruby-minimum change. Distinct values no longer collapse into one Hash key. Nil-modseq Array compatibility and the non-nil same-class restriction remain. Plain Array receivers still use Array equality; this does not promise cross-class symmetry for arbitrary Array subclasses. Other Ruby/OS versions were not run locally. No production or external IMAP service was used. Local verification does not imply upstream CI approval or comprehensive behavioral coverage.

### DIFF
--- a/lib/net/imap/search_result.rb
+++ b/lib/net/imap/search_result.rb
@@ -80,8 +80,12 @@ module Net
 
       # Hash equality.  Unlike #==, order will be taken into account.
       def eql?(other)
-        return super if modseq.nil?
-        self.class == other.class && hash == other.hash
+        return false unless super
+        if modseq.nil?
+          !other.respond_to?(:modseq) || other.modseq.nil?
+        else
+          self.class == other.class && modseq == other.modseq
+        end
       end
 
       # Returns a string that represents the SearchResult.
--- a/lib/net/imap/search_result.rb
+++ b/lib/net/imap/search_result.rb
@@ -73,19 +73,13 @@ module Net
       end
 
       # Hash equality.  Unlike #==, order will be taken into account.
-      def hash
-        return super if modseq.nil?
-        [super, self.class, modseq].hash
-      end
+      def hash = [super, self.class, modseq].hash
 
       # Hash equality.  Unlike #==, order will be taken into account.
       def eql?(other)
-        return false unless super
-        if modseq.nil?
-          !other.respond_to?(:modseq) || other.modseq.nil?
-        else
-          self.class == other.class && modseq == other.modseq
-        end
+        self.class == other.class &&
+          modseq == other.modseq &&
+          super
       end
 
       # Returns a string that represents the SearchResult.
--- a/test/net/imap/test_search_result.rb
+++ b/test/net/imap/test_search_result.rb
@@ -30,6 +30,17 @@ class SearchDataTests < Net::IMAP::TestCase
     refute_equal nomodseq, sorted
   end
 
+  test "#eql? and #hash include the result type and modseq" do
+    result = SearchResult[1, 2, modseq: 3]
+    same = SearchResult[1, 2, modseq: 3]
+
+    assert_operator result, :eql?, same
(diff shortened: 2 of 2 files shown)


lib/net/imap/search_result.rb
```diff
@@ -80,8 +80,12 @@ def hash
...
       # Hash equality.  Unlike #==, order will be taken into account.
       def eql?(other)
-        return super if modseq.nil?
-        self.class == other.class && hash == other.hash
+        return false unless super
+        if modseq.nil?
+          !other.respond_to?(:modseq) || other.modseq.nil?
+        else
+          self.class == other.class && modseq == other.modseq
+        end
```
nevans: `return false unless super` is good, except that checking modseq could be _much_ faster than checking `Array#eql?`, and it's a good idea to short-circuit when possible.  So the modseq should be checked first, and call `super` last:

```suggestion
        if modseq.nil?
          !other.respond_to?(:modseq) || other.modseq.nil?
        else
          self.class == other.class && modseq == other.modseq
        end &&
          super
```

I think we should use the same type comparison on both sides of the conditional, either `other.respond_to?(:modseq)` or `self.class == other.class`.  I'd also be okay with `self.class === other`.  But I lean against using `respond_to?` in `eql?`.

I'll note that `SearchResult#==` also uses `respond_to?(:modseq)`, but it's normal for `#eql?` to be stricter than `#==`.

```suggestion
        if modseq.nil?
          !(self.class === other) || other.modseq.nil?
        else
          self.class === other && modseq == other.modseq
        end &&
          super
```

That can be further simplifed as
```suggestion
        (self.class === other ? modseq == other.modseq : modseq.nil?) &&
          super
```

Looking at `SearchResult#==`, it has an even nicer way to compare modseq:
```suggestion
        modseq == (other.modseq if self.class === other) &&
          super
```

As a bugfix, I'd be happy with that.  But, as a backward incompatible change (for 0.7.0), I'd prefer to go further:

```suggestion
        self.class === other &&
          modseq == other.modseq &&
          super
```

nevans (changes requested):
I remember finding #514.  It's too bad I didn't spot this at the same time!

Can you add a few tests please?

nevans:
Related: `SearchResult#hash` should _always_ incorporate `modseq` into its hash, not only when it's not `nil`.  While that wouldn't 100% fix the issues you identified, it _would_ make them far less likely (for hash keys... but `#eql?` is used for a lot more than just hash keys).

```ruby
def hash = [super, self.class, modseq].hash
```

nevans (approved):

PR #740, opened 2026-08-27
Match ID string literal bytes without encoding conflicts

OskarEichler:
## Summary

Match the literal-selection regexp against a binary view of ID strings. Matching its high-byte range directly against UTF-8 raises Encoding::CompatibilityError before ClientID can send ordinary non-ASCII names or values. Preserve the original bytes and the existing quoted-versus-literal policy.

## Reproduction

```ruby
require 'net/imap'
value = Net::IMAP::StringFormatter.string('日本語')
p [value.class, value.data.bytesize] # [Net::IMAP::Literal, 9]
# Before: Encoding::CompatibilityError.
```

## Verification

- 92 focused checks; 22 failing expectations before, zero afterward. UTF-8, ISO-8859-1, binary and invalid UTF-8 byte strings, frozen/mutable inputs, ASCII quoting, CR/LF, nil and symbols, plus ClientID key/value formatting are covered.
- Existing `rake test` on this isolated branch: **1726 tests, 12594 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications**, Ruby 4.0.6 via rbenv. The baseline also passes all 1,726 tests; assertion counts vary slightly across runs.

## Compatibility and limits

No signature or wire-format change for previously working input. Non-ASCII strings are sent as literals without transcoding. This is a formatting fix, not a change to charset negotiation or ID input validation. Other Ruby/OS versions were not run locally. No production or external IMAP service was used. Local verification does not imply upstream CI approval or comprehensive behavioral coverage.

### DIFF
--- a/lib/net/imap/command_data.rb
+++ b/lib/net/imap/command_data.rb
@@ -459,7 +459,7 @@ module Net
       # coerces using +to_s+
       def string(str)
         str = str.to_s
-        if str =~ LITERAL_REGEX
+        if str.b.match?(LITERAL_REGEX)
           Literal.new(str)
         else
           QuotedString.new(str)
--- a/test/net/imap/test_command_data.rb
+++ b/test/net/imap/test_command_data.rb
@@ -161,6 +161,14 @@ class CommandDataTest < Net::IMAP::TestCase
   class StringFormatterTest < Net::IMAP::TestCase
     include Net::IMAP::StringFormatter
 
+    test "string selects a literal for non-ASCII text" do
+      value = "日本語"
+      formatted = string(value)
+
+      assert_kind_of Literal, formatted
+      assert_equal value.b, formatted.data
+    end
+
     test "literal_or_literal8" do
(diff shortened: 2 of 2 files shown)


nevans:
I have some other work in progress that will conflict with (clobber) this.  But I'll either merge yours first (so it can go out in the next release) or merge mine into this PR (and give you the credit 😉).

nevans (changes requested):
Even though I'll probably clobber your implementation with mine, I might not finish mine before the next release.  Your bugfix is tiny and focused, and fixes the bug.  Mine is part of a bigger refactoring of how `Net::IMAP` handles arguments and compiles commands.

So, can you add a small test for this, so it can go out with the next release?  Thanks!

OskarEichler:
Added a focused regression test covering non-ASCII text through StringFormatter.string. The full command-data test file passes: 59 tests, 427 assertions.

nevans (approved):
Thanks!

PR #741, opened 2026-08-27
Accept string arguments in SequenceSet delete?

OskarEichler:
## Summary

Use String#match? when deciding whether a delete? argument represents one number or a range. String#include? does not accept a Regexp, so documented string inputs currently raise TypeError. Retain the numeric-versus-SequenceSet return distinction introduced by #583.

## Reproduction

```ruby
require 'net/imap'
set = Net::IMAP::SequenceSet.new('1:5')
p set.delete?('2') # 2
p set.delete?('3:4').to_s # "3:4"
p set.to_s # "1,5"
# Before: TypeError on the first deletion.
```

## Verification

- 39 focused checks; 28 failing expectations before, zero afterward. Empty/nonempty sets, present/absent numbers, reversed/equal ranges, star and frozen rejection are covered. The source of merged #583 and its earlier numeric/range return policy was reviewed.
- Existing `rake test` on this isolated branch: **1726 tests, 12604 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications**, Ruby 4.0.6 via rbenv. The baseline also passes all 1,726 tests; assertion counts vary slightly across runs.

## Compatibility and limits

No public signature change. String numbers now return a number and string ranges return a SequenceSet, as intended. Missing entries still return nil and frozen sets still raise FrozenError. Other Ruby/OS versions were not run locally. No production or external IMAP service was used. Local verification does not imply upstream CI approval or comprehensive behavioral coverage.

### DIFF
--- a/lib/net/imap/sequence_set.rb
+++ b/lib/net/imap/sequence_set.rb
@@ -1922,7 +1922,7 @@ module Net
       def number_input?(input)
         case input
         when *STARS, Integer then true
-        when String          then !input.include?(/[:,]/)
+        when String          then !input.match?(/[:,]/)
         end
       end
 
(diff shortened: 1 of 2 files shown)


nevans (changes requested):
please add a test

nevans (approved):

PR #742, opened 2026-08-27
Handle zero and negative SequenceSet min/max counts

OskarEichler:
## Summary

Return an empty SequenceSet for max(0), reject negative min/max counts, and normalize integer-like counts before slicing. max(0) currently slices from -0 and returns the whole set; negative counts produce unrelated slices. Preserve oversized-count and frozen-result behavior.

## Reproduction

```ruby
require 'net/imap'
set = Net::IMAP::SequenceSet['1,3,5']
p set.max(0).to_s # before: "1,3,5"; after: ""
begin
  set.min(-1)
rescue ArgumentError => e
  p e.message # "negative count"
end
```

## Verification

- 200 focused checks; 34 failing expectations before, zero afterward. Empty/mutable/frozen/star sets, zero/positive/oversized/negative counts and objects implementing to_int are covered. Existing #460 defines count behavior and #580 fixes a separate cardinality/duplicate case already present in master.
- Existing `rake test` on this isolated branch: **1726 tests, 12598 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications**, Ruby 4.0.6 via rbenv. The baseline also passes all 1,726 tests; assertion counts vary slightly across runs.

## Compatibility and limits

Intentional correction: negative counts now raise ArgumentError rather than returning a slice. Zero yields an empty set. Omitted/nil/false count behavior and scalar star handling are unchanged. Integer-like counts use the existing SequenceSet to_int coercion convention. Other Ruby/OS versions were not run locally. No production or external IMAP service was used. Local verification does not imply upstream CI approval or comprehensive behavioral coverage.

### DIFF
--- a/lib/net/imap/sequence_set.rb
+++ b/lib/net/imap/sequence_set.rb
@@ -788,6 +788,9 @@ module Net
       # Related: #min, #minmax, #slice
       def max(count = nil, star: :*)
         if count
+          count = Integer(count.to_int)
+          raise ArgumentError, 'negative count' if count < 0
+          return remain_frozen_empty if count == 0
           if cardinality <= count
             frozen? ? self : dup
           else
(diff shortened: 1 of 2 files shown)


nevans (changes requested):
please add tests

nevans (approved):

nevans (approved):

PR #748, opened 2026-08-27
Parse STATUS responses with empty attribute lists

OskarEichler:
## Summary

Accept an empty attribute list in a STATUS response and return StatusData with an empty attr hash. The response grammar permits the list to be absent inside parentheses, but the parser currently unconditionally requires its first attribute.

## Reproduction

```ruby
require 'net/imap'
p Net::IMAP::ResponseParser.new.parse("* STATUS INBOX ()\r\n").data.attr
# Before: ResponseParseError. After: {}.
```

## Verification

- 74 focused checks, 48 failing expectations before and zero afterward. Atom/quoted/literal mailbox names, keyword casing, existing trailing-space tolerance, frozen inputs, input retention and populated responses are covered. RFC3501 section 9 explicitly makes status-att-list optional: https://www.rfc-editor.org/rfc/rfc3501.html#section-9
- Existing `rake test` on this isolated branch: **1726 tests, 12594 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications**, Ruby 4.0.6 via rbenv. Baseline also passes 1,726 tests; assertion counts vary slightly between runs.

## Compatibility and limits

No signature or existing valid-response shape change. Previously rejected valid empty responses now return an empty hash. Nonempty attribute parsing is unchanged; no new malformed-input tolerance is introduced. No production or external IMAP service used. Other Ruby/OS versions were not run locally. Local success does not imply upstream CI approval or exhaustive coverage.

### DIFF
--- a/lib/net/imap/response_parser.rb
+++ b/lib/net/imap/response_parser.rb
@@ -1676,7 +1676,7 @@ module Net
       def mailbox_data__status
         resp_name  = label("STATUS"); SP!
         mbox_name  = mailbox;         SP!
-        lpar; attr = status_att_list; rpar
+        lpar; attr = peek_rpar? ? {} : status_att_list; rpar
         UntaggedResponse.new(resp_name, StatusData.new(mbox_name, attr), @str)
       end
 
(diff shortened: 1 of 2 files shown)


nevans (changes requested):
Can you add an example to `test/net/imap/fixtures/response_parser/status_responses.yml`, please?

If you leave out the `:expected:` value, the generated test will be pending and it will print out the appropriate `:expected:` for you to inspect, copy, and paste.

OskarEichler:
Added the requested status_responses.yml fixture for an empty STATUS attribute list. The full response-parser test file passes: 211 tests, 281 assertions.

nevans (approved):
Thanks!

PR #749, opened 2026-08-27
Parse NIL APPENDLIMIT values in STATUS responses

OskarEichler:
## Summary

Handle the NIL alternative of STATUS APPENDLIMIT explicitly. Numeric APPENDLIMIT already parses through generic number handling, but a mailbox reporting no limit currently raises ResponseParseError. Preserve numeric values and represent NIL as Ruby nil.

## Reproduction

```ruby
require 'net/imap'
p Net::IMAP::ResponseParser.new.parse("* STATUS INBOX (APPENDLIMIT NIL)\r\n").data.attr
# Before: ResponseParseError. After: {"APPENDLIMIT" => nil}.
```

## Verification

- 60 focused checks, 24 failing expectations before and zero afterward. Keyword/NIL casing, zero/ordinary/maximum 32-bit numeric values, frozen inputs and surrounding attributes are covered. RFC7889 sections 3 and 5 specify NIL for no mailbox limit: https://www.rfc-editor.org/rfc/rfc7889.html#section-5
- Existing `rake test` on this isolated branch: **1726 tests, 12602 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications**, Ruby 4.0.6 via rbenv. Baseline also passes 1,726 tests; assertion counts vary slightly between runs.

## Compatibility and limits

No signature, dependency or Ruby-minimum change. Valid APPENDLIMIT NIL responses now parse; numeric values retain Integer representation. This does not implement client-side APPEND size enforcement or change unknown extension parsing. No production or external IMAP service used. Other Ruby/OS versions were not run locally. Local success does not imply upstream CI approval or exhaustive coverage.

### DIFF
--- a/lib/net/imap/response_parser.rb
+++ b/lib/net/imap/response_parser.rb
@@ -1743,6 +1743,7 @@ module Net
           when "SIZE"          then number64            # RFC8483, RFC9051
           when "HIGHESTMODSEQ" then mod_sequence_valzer # RFC7162
           when "MAILBOXID"     then parens__objectid    # RFC8474
+          when "APPENDLIMIT"   then NIL? ? nil : number  # RFC7889
           else
             number? || ExtensionData.new(tagged_ext_val)
           end
--- a/test/net/imap/fixtures/response_parser/status_responses.yml
+++ b/test/net/imap/fixtures/response_parser/status_responses.yml
@@ -25,6 +25,16 @@
           UIDVALIDITY: 1234
       raw_data: "* STATUS INBOX (UIDNEXT 1 UIDVALIDITY 1234)\r\n"
 
+  test_status_response_appendlimit_nil:
+    :response: "* STATUS INBOX (APPENDLIMIT NIL)\r\n"
+    :expected: !ruby/struct:Net::IMAP::UntaggedResponse
+      name: STATUS
+      data: !ruby/struct:Net::IMAP::StatusData
+        mailbox: INBOX
+        attr:
+          APPENDLIMIT:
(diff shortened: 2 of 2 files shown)


nevans (changes requested):
Similar to #748, can you add an example to `test/net/imap/fixtures/response_parser/status_responses.yml`, please?

If you leave out the `:expected:` value, the generated test will be pending and it will print out the appropriate `:expected:` for you to inspect, copy, and paste.

OskarEichler:
Added the requested status_responses.yml fixture for APPENDLIMIT NIL. The full response-parser test file passes: 211 tests, 281 assertions.

nevans (dismissed):
Thanks!

OskarEichler:
Upstream added the adjacent empty-STATUS fixture after the approval, which created a content conflict in the same YAML section. I merged current master, retained both fixtures without changing the APPENDLIMIT parser fix, and reran the focused response-parser suite: 212 tests, 282 assertions, all passing. GitHub dismissed the approval because of the merge commit, so this is ready for a quick re-approval when convenient.

nevans (approved):

nevans:
I mislabeled this as a bug earlier.  But the `APPENDLIMIT` extension does not fit into the the formal syntax of [RFC3501](https://www.rfc-editor.org/info/rfc3501/), [RFC4466](https://www.rfc-editor.org/info/rfc4466/), or [RFC9051](https://www.rfc-editor.org/info/rfc9051/), so this adds support for parsing the `APPENDLIMIT` extension to status responses.

PR #769, opened 2026-09-12
🐛 Fix ResponseParser use of global config

nevans:
ResponseParser was using the global config here, but should've used its own instance config.


nevans:
I thought I'd merged this already, before v0.6.7. 😞

